PR #1, opened 2016-04-26
Add astUnitNormMap

timj:
Now that AST is standalone I've ported across @r-owen's patches from timj/starlink-pyast#13 to AST and added a Fortran wrapper.

I've also "fixed" spelling in a separate commit. If we are happy with "center" instead of "centre" then we can remove that commit.


r-owen:
Thank you very much. I think centre is probably better for consistency; sorry I missed that earlier.

dsberry:
Consistency in the docs (and code) is good. The other issue is attribute names - not currently a problem for UnitNormMap, but it is possible to allow alternate spellings for attributes in AST. For instance the Plot class "Colour" attribute can also be specified as "Color".  Maintaining this sort of bilingual facility is a good idea as attribute names are a run-time issue rather than just a documentation issue, and could potentially be supplied by users (as in Starlink graphics style files). But in general, life is easier if we try to stick to names with no transatlantic differences, maybe by using synonyms or abbreviations ("e.g."PcdCen" rather than "PcdCentre/er"). Although of course that's not always possible.

PR #2, opened 2016-10-12
Prevent PGPLOT interface from being built if Fortran is disabled

timj:
The problem is that f77.h can't be built if Fortran is disabled
but the PGPLOT grf interface assumes that f77.h is available.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -258,6 +258,24 @@ F_C_INCLUDE_FILES =
 GRP_F_INCLUDE_FILES =
 endif
 
+# If we have no Fortran we are not building f77.h and we probably
+# do not want a Fortran runtime. This requires that PGPLOT is disabled.
+# We replace it with the stub GRF interface. An alternative would
+# be to have the PGPLOT wrappers use a preprocessor symbol to build
+# the pgplot to always error if used.
+if !NOFORTRAN
+GRF_PGPLOT_SOURCES = \
+    grf_pgplot.c
+GRF3D_PGPLOT_SOURCES = \
+    grf3d_pgplot.c
+else
+GRF_PGPLOT_SOURCES = \
+    grf_5.6.c
+GRF3D_PGPLOT_SOURCES = \
+    grf3d.c
+endif
+
+
 ## Following declaration isn't used
 ## LATEX_DOCUMENTATION_FILES = \
 ##     sun210.tex \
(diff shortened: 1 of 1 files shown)


timj:
@dsberry can you take a look at this? It might not be the best approach but we need it to get a non-Starlink build to work with Fortran disabled. It might be better for `--without-fortran` to set a CPP directive that prevents `f77.h` from being loaded and instead defines stub interfaces that complain if used.

Can I ask why AST isn't using the actual C PGPLOT interface? Switching to that would be the easier fix.

dsberry:
AST uses its own F77 interface rather than the caltech C interface for two reasons (as far as I can remember): 1) so that the GKS-based starlink version of pgplot (which does not have a C interface) can be used if required, and 2) because the grf3d_pgplot.c file calls some internal  pgplot subroutines that are not included in the caltech C interface.

I think it would be perfectly safe now to drop support for the starlink version of pgplot.

The grf3d issue is more of a problem. It calls the internal pgplot functions that provide the polygons defining each text character. It needs to do this so it can transform the vertices into 3D and so display text in a projected manner.

I think my preferred solution at the moment would be to drop support for starlink pgplot, change grf_pgplot.c to use the caltech interface, and then adopt your solution just for grf3d_pgplot.c. Does that sound OK to you?

timj:
That all sounds fine, but in the interim is it okay if we merge this patch in its current state as it should only affect non-Starlink users? We can't build the LSST fork without this fix and I assume you are not proposing to fix `grf_pgplot.c` today? Obviously, it would be nice if people were told at runtime if their PGPLOT interface has been replaced by the stub but I think this patch is the minimum required to allow us to make progress here. @r-owen is working on integrating AST into the LSST build system.

timj:
Thanks. I see this was merged as a fast forward rather than a merge commit. Did you use the default option on the pull request to do that or did you select the squash option on the pull down?

dsberry:
I selected the "Rebase and merge" option. I thought that was what we usually did, although I'm guessing you're about to tell it it's not...

timj:
Regarding the `pgp` problem. The "proper" fix would seem to be to copy the Fortran interface code over to the `pgp` package itself rather than having that code in the AST library. I understand that it's hard to find motivation for that given the deprecated status of GKS PGPLOT.

timj:
"Rebase and merge" is the new (I would say) broken Github scheme for rebasing. I always prefer a "rebase and do not fast forward when merging" scheme because it shows you what commits are contributing to the new feature. Github's fast forward approach results in some commits appearing on master without any idea of where they came from and how they relate to each other. In this case it's a single commit so it can stand alone. Not a problem, I just wondered if Github had changed the default behavior.

PR #3, opened 2017-08-01
Fix UnitNormMap transformation of the centre

r-owen:
Modify UnitNormMap to transform the centre properly, and in a way that round trips:
- In the forward direction transform centre to [AST__BAD, ..., AST__BAD, 0]
  (since the components of the unit vector are unknown, but the norm is 0)
- In the inverse direction transform [any_value, ..., any_value, 0] to [0, ..., 0]

Formerly UnitNormMap transform input = centre to [AST__BAD, ..., AST__BAD],
which was incorrect and did not round trip.


dsberry:
Merged.

PR #32, opened 2026-02-11
Some minor fixes found from pyast

timj:
* The compiler was complaining that AST__KINT was indexing into a static array that wasn't big enough.
* Fixed problem stringifying a float which resulted in a leading space being added.

See Starlink/starlink-pyast#45

Dropped this change:

* The PRId64 definition is gated on a configure test that pyast wasn't doing. Given that AST__DIMFMT existed and is used in some other places, I was wondering if it made sense to use it in fitschan/moc/fitstable as well.

following discussion.

### DIFF
--- a/src/fitschan.c
+++ b/src/fitschan.c
@@ -26488,8 +26488,13 @@ static void RoundFString( char *text, int width, int fitsrnd, int *status ){
          }
       }
 
-/* Copy the rounded string into the supplied text string, if there is room. */
-      if( astChrLen( ltext ) <= len0 ) strcpy( text, ltext );
+/* Copy the rounded string into the supplied text string, if there is room,
+   correcting for the offset thta was added at the start of this function if
+   there was no rounding. */
+      c = ltext;
+      if( *c == ' ' ) c++;
+      if( astChrLen( c ) <= len0 ) strcpy( text, c );
+
 
 /* Free local resources. */
       ltext = astFree( ltext );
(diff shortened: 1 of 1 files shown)


timj:
It looks like the change in the leading space is down to the rewrite of `RoundFString`.

dsberry (commented):
I seem to have introduced an inconsistency in the way AST handles 8 byte ints. It tries to cater for compilers that do not have an int64_t type by defining AstDim as "long int" or "long long int" if int64_t is not available (see object.h). Yet at the same time it uses "int64_t" explicitly at various points in the code, thus preventing compilation by compilers that do not have int64_t. It seems to be these explicit uses of int64_t that are associated with the use of PRId64 instead of AST__DIMFMT.  The long and complicated solution is probably to remove every explicit use of int64_t (and PRId64) in AST. A simpler and more pragmatic solution may be to drop support for systems that do not have int64_t (i.e. do away with AstDim and just use int64_t instead). Simply sticking with the changes you have made would also work of course, since AstDim *is* int64_t (and AST__DIMFMT is PRId64) but is probably not as tidy and clear.

dsberry:
The other two look fine. The test suite passes with your changes (I have just updated the SplineMap C tester since I failed to update it to take account of commit  7be740704 - "Ensure good values are returned for input points on the top edge of the know bounding box"). If pyast can easily be made to work without the PRId64 changes, then maybe leave them out for now until such time as all the AstDims can be replaced by int64_t.

timj:
Thanks. Okay. I will drop the PRId64 change. I ended up setting HAS_INT64_T explicitly in the pyast build so that should fix it for me.

PR #33, opened 2026-02-24
Remove the GPL files

timj:
The library is LGPL since bddbbf7 but incongruously that also added GPL license to the repository.

@dsberry I'm a bit confused about this situation. Do you have any clarity regarding why GPL and LGPL were added at the same time?


dsberry (approved):
I haven't a clue now why the GPL files were added. Removing them seems perfectly reasonable.

PR #35, opened 2026-03-05
yamlchan: Add spherical_cartesian support

embray:
Greetings, allow me to introduce myself with a small pull request.  I've been tasked by STScI to evaluate the suitability of AST for evaluating the WCS of Roman Space Telescope imaging data, and so far I'm extremely impressed and encouraged.

It seems the `YamlChan` implementation has not been updated in a few years, and ASDF + GWCS have evolved a bit since then, but not drastically.  This adds support for the `spherical_cartesian` transform documented at https://www.asdf-format.org/projects/asdf-wcs-schemas/en/latest/generated/gwcs/spherical_cartesian-1.1.0.html

I didn't add support yet for the optional `wrap_lon_at` property, but that would be an easy-enough extension. I also tried my best to understand and mimic your coding style.

This also updates the supported minor versions of some of the existing transforms already supported in the library.  I haven't gone through all those transforms yet to see what changed--if they have new features that should be supported.  But only the minor versions changed so they should still be backwards-compatible.

I have a couple additional patches I'm still working on but this already looks to have 90% of what I'm aware will be needed for RST.

----

As a side-discussion (that maybe we can take elsewhere), this is part of my larger work on [libasdf](https://github.com/asdf-format/libasdf), a native C library for ASDF.  I wonder if there would be any interest in, or even any reason to integrate it into AST.  libasdf is built around the newer libfyaml (as opposed to libyaml), and has full support for reading the binary block data, including compressed.

On one hand, the code in `YamlChan` is already quite stable and probably doesn't need to be disrupted.  AST already has its own native `AstKeyMap` data structure, so in terms of handling the YAML structure libasdf wouldn't add much.

The support for binary blocks might be more useful, as this is a current limitation.  However, as WCS is concerned this only shows up in a few transforms--in particular polynomial transforms store their coefficients in ndarray.  The ASDF ndarray type supports both binary data, and in-line data in the YAML tree.  Personally, I think it's a bit silly that GWCS tends to always store polynomial ndarray's with data stored as binary instead of inline, as these are typically just small matrices; that's an issue I plan to raise upstream with GWCS.  In the meantime, I'm working around this with a preprocessing step that converts the binary ndarrays back to pure YAML.

But maybe a first-class `AsdfChan` would also be useful for ASDF write support.  I'm just not sure exactly where that might fit in, if at all, or how difficult it would be to refactor with reuse of the existing YamlChan support code.


timj:
Thank you for this. It looks good to me but I'll give @dsberry a chance to take a look.

I have been pondering the GitHub Action situation since it's obviously less than ideal to have no build test phase.

I did recently make a new release of starlink-pyast which does now publish wheels to PyPI and includes tests so I am making some progress. We are wanting to make use of pyast at Rubin so that Astropy users without access to the LSST Science Pipelines can access the Rubin transforms.

I had meant to contact the ASDF team at some point because Rubin has recently added SplineMap to AST for fine-grained WCS corrections and we realized that would mean we couldn't use the YAML ASDF serialization.

I spoke to Nadia at Malta ADASS and I was hoping that libasdf would be written in Rust but it seems like you chose C?

embray:
Hi @timj, thanks a bunch for having a look!

Yes, I noticed that about the CI.  I haven't even run the tests for AST yet because I couldn't get the fortran build working :)  Something I need to look at again.

Interesting about SplineMap.  I hadn't seen it yet.  Is there something it does that's not compatible with the Astropy spline model?

> I spoke to Nadia at Malta ADASS and I was hoping that libasdf would be written in Rust but it seems like you chose C?

When we first kicked off the project this was discussed, and I proposed the idea of writing it in Rust.  We ended up sticking with C for maximum interoperability across the ecosystem with minimum friction--basically the lowest-common denominator.  And Rust as a dependency is still uneven compared to "just link against an .so".  Already had some success integrating libasdf into, e.g. SourceXtractor++ which would've been possible with Rust but with a lot more hassle.

That said, it would still be a good idea to have, and if there's demand from the community I think a Rust crate wrapping the C library would be a good start.

dsberry:
Yes, thanks for this.  It's nice to see more people getting involved in AST development.

 I'm a bit confused by the changes to the minor versions given that you say you've not checked what has actually changed in the affected classes. Is there any chance that there may be changes that are not backward compatible and so may cause YamlChan to misinterpret the object? I'm guessing that since it is only the minor version that has changed, all changes will be backward compatible and shouldn't cause any problems if they are simply ignored?

Also, there seems to be no support for writing (i.e. converting an AST SphMap to a YAML-encoded spherical_cartesian). 

In the IsA function (line 4715) you compare the class name to "spherical_cartesian" and also "transform" before checking if the object is a spherical_cartesian. Why do you compare the name to "transform"?

Should the IsATransform function be modified to include a test for a spherical_cartesian?

The comments describing the YamlEncoding attribute (line 18008) should probably be changed to include the new class.

To be consistent with the rest of the code, you should probably declare the IsASpherical_Cartesian function using the MAKE_PROTO macro (see line 516) rather than explicitly as you do at line 4646.

In ReadSphericalCartesian, the variables cm and cm2 are unused.

Regarding testing, I agree that tests being written in Fortran is now a problem. It was done originally because AST was mainly used from Fortran in the early days, and testing in Fortran tested both the C code and the Fortran wrappers. We could really do with a test suite that doesn't require Fortran.

src/yamlchan.c
```diff
@@ -4708,6 +4712,8 @@ static int IsA( AstKeyMap *km, const char *class, int *status ) {
             result = IsAFrame( km_class, status );
          } else if( !strcmp( "frame2d", class ) ){
             result = IsAFrame2d( km_class, status );
+         } else if( !strcmp( "spherical_cartesian", class ) || !strcmp( "transform", class ) ){
+            result = IsASpherical_Cartesian( km_class, status );
```
embray, line 4863: Doesn't make sense here at all.  It should be moved to the `if( !strncmp( km_class, "asdf/transform/", 15 )` below which I think I was overlooking.

embray:
> Is there any chance that there may be changes that are not backward compatible and so may cause YamlChan to misinterpret the object? I'm guessing that since it is only the minor version that has changed, all changes will be backward compatible and shouldn't cause any problems if they are simply ignored?

It's a good question.  I'll take the time to go through what actually changed in those in a bit, and maybe split it to a separate PR.  In any case the versions need to be updated or else AST refuses to read the newer versions of these transforms found in the Roman files.

> The comments describing the YamlEncoding attribute (line 18008) should probably be changed to include the new class.

Yep

> Should the IsATransform function be modified to include a test for a spherical_cartesian?

I guess it should be; I overlooked that.

> To be consistent with the rest of the code, you should probably declare the IsASpherical_Cartesian function using the MAKE_PROTO macro (see line 516) rather than explicitly as you do at line 4646.

Ah, right. I meant to fix that.

> In ReadSphericalCartesian, the variables cm and cm2 are unused.

Oops, left over from a previous attempt.

> Also, there seems to be no support for writing (i.e. converting an AST SphMap to a YAML-encoded spherical_cartesian).

I admit I have not even looked at the write support yet.  I'm just trying to get AST to read Roman WCS's.  I'll take a look; it should be made to work.

src/yamlchan.c
embray: Actually, I remember what I was doing here now.  It does belong here because `spherical_cartesian` is under the `stsci.edu/gwcs` namespace.  It's because this is still specific to the gwcs package: https://github.com/spacetelescope/gwcs/blob/e87b1d5ceb220b2ee714331d208f7f69e08a4991/gwcs/geometry.py#L68

embray: This still seems wrong actually.  I keep getting confused here between `km_class` and `class`.  I'll get this straightened out.

embray: I think what I had here before was actually right--the point is that there can be some subclasses of "transform" in the "gwcs/" namespace as well as in the "asdf/transform/" namespace.  It just happens that this is the only GWCS-specific transform implemented so far, but I'll try to make that a bit clearer.

embray:
Added write support, round-trippable, for spherical_cartesian.  It basically follows the same pattern as with `FindRotate3d` and `FindAffine`.  It makes me wonder if when reading out of the ASDF file it wouldn't make sense in the first place to take the CmpMap returned for the spherical_cartesian and add a property on it noting that it was serialized from an ASDF spherical_cartesian transform.  That would allow skipping this "find" step, though it's only really useful for round-tripping transforms read from an existing file.

I have a test program (written in C) for this as well, but I'm not really sure where best for it to land. I have some thoughts on that though coming soon in another PR...

dsberry:
As you say, we need the "find" step anyway, in order to cope with cases where the AST SphMap was not read from an ASDF spherical_cartesian transform.

I am in the process of converting the main yamlchan tester from Fortran to C, so your new testing code would nicely fit in there. I'll let you know when I've completed the conversion.

dsberry:
Your code looks good. I'm impressed that you are clearly getting your head round the inwards of AST! The only point is that the way you have written WriteSphMap prevents correct round-tripping (AST -> ASDF -> AST) for a simple SphMap. I think it should explicitly take account of the difference in angle units (radians versus degrees) between AST and ASDF. Otherwise the following test fails:

```
#include "ast.h"
#include <stdio.h>

int main(){

   AstFrame *frm1;
   AstFrame *frm2;
   AstFrameSet *fs;
   AstObject *obj;
   AstSphMap *map;
   AstYamlChan *ch;

   astBegin;

/* Create the YamlCHan, telling it to write text to disk file "aa.asdf" */
   ch = astYamlChan( NULL, NULL, "SinkFile=aa.asdf" );

/* Create a simple FrameSet that uses a SphMap to convert 3-D cartesian
   to 2-D spherical coords. */
   frm1 = astFrame( 3, "Domain=Cartesian" );
   frm2 = astFrame( 2, "Domain=Spherical,Symbol(1)=Alpha,Symbol(2)=Beta" );
   map = astSphMap( " " );
   fs = astFrameSet( frm1, " " );
   astAddFrame( fs, AST__BASE, map, frm2 );

/* Attempt to write out the FrameSet through the YamlChan. */
   if( astWrite( ch, fs ) != 1 ) {
      printf("Failed to write FrameSet to YamlChan\n");

/* If successful, force the YamlChan to write its contents to disk by
   clearing the SinkFile attribute. Then read the disk file back again
   and attempt to read an AST object (a FrameSet) from it. */
   } else {
      astClear( ch, "SinkFile" );
      astSet( ch, "SourceFile=aa.asdf" );
      obj = astRead( ch );

/* Display the Mapping from Base to Current Frame. For correct
   round-tripping, this should be a simple SphMap. */
      astShow( astGetMapping( obj, AST__BASE, AST__CURRENT ) );
   }

   astEnd;
}
```
The  best place to do this may be the WriteAsdfStep function, which already takes account of the (radians,degrees) issue when writing out celestial coordinate frames. A possible approach may be to check if the supplied Mapping is a SphMap (using astIsASphMap) and, if so, add in a suitable ZoomMap to convert the units (using AddR2D).

dsberry:
Also, I've committed the change to convert the yamlchan tester from Fortran to C.

dsberry:
Having thought a bit more about this, I think I was wrong to suggest changing WriteASDFStep, because it would only work for SphMaps that formed the entire mapping between steps. It would not work for a SphMap contained within a compound mapping. A better idea may be to change WriteSphMap so that it uses AddR2D to add a radian/degree conversion to the transform's KeyMap, before or after the spherical_cartesian.

embray:
You're right -- I was only really thinking about the case of round-tripping from ASDF, but didn't really handle the case of a hand-constructed AST map containing a SphMap in it somewhere.  Your proposal to then couch it in the appropriate unit conversions makes sense.  I'll give that a try.

embray:
I think this handles the round-tripping correctly now ; I also added the test in testyamlchan.c (which I cleaned up a bit so that the individual tests are clearer).

In my test I neglected to provide names for the axes but that exposed an annoying bug -- since the GWCS format doesn't strictly require specifying the axes in a given frame in most cases, the naxes is basically implicit based on the transform (some frames only allow a specific number of axes, but most don't actually require providing `axes_names`, `axis_physical_types` etc.).  This leads to errors in creating the frame in the final step, because there is no way to determine without additional hinting what the expected naxes is for the final frame.

The solution is to simply pass a hint through from each `ReadStep` call, using the Nout of the previous step's transform as the `nax_hint`.

The file generated by the test now looks like this:

```yaml
#ASDF 1.0.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
wcs: !<tag:stsci.edu:gwcs/wcs-1.0.0>
  name: ''
  steps:
  - !<tag:stsci.edu:gwcs/step-1.0.0>
    frame: !<tag:stsci.edu:gwcs/frame-1.0.0>
      name: cart3d
    transform: !transform/compose-1.2.0
      forward:
      - !<tag:stsci.edu:gwcs/spherical_cartesian-1.3.0>
        transform_type: cartesian_to_spherical
      - !transform/concatenate-1.2.0
        name: AST__DD2R
        forward:
        - !transform/scale-1.2.0
          name: AST__DD2R
          factor: 0.017453292519943295
        - !transform/scale-1.2.0
          name: AST__DD2R
          factor: 0.017453292519943295
  - !<tag:stsci.edu:gwcs/step-1.0.0>
    frame: !<tag:stsci.edu:gwcs/frame2d-1.0.0>
      name: sph
...
```

dsberry:
@embray That all looks good. Only point is that the description in the prologue of WriteSphMap still contains " Note: the ASDF spherical_cartesian schema uses degrees " etc that presumably is now no longer relevant.

Do you still have further work to do on this PR? If not, I'm happy to rebase and merge. The only thing I'd ask is that you add suitable entries to the "Authors:" and "History:" sections in the main prologue at the top of yamlchan.c.

embray:
@dsberry Will do.  I have a follow-up PR coming that will add support for another new transform that shows up in the RST calibration file WCS, but easier to get this in first since there will likely be merge conflicts.  I think it's a bit simpler one.

PR #37, opened 2026-04-01
Out-of-Starlink build support including running Fortran tests and basic CI

embray:
### Summary

This PR does a few things:

- Supports building AST without the Starlink build system and library ecosystem
- Also supports building and running the Fortran tests in the standalone build -- this requires cherry-picking a small amount of code from Starlink into this repository though
- The tests can also be run as a standard automake test suite with `make check`, rather than going through 
- Adds a very simple github actions workflow for building and testing on Linux and macOS

### Details

I'm always wary of touching someone else's build system, but I tried to keep this as non-invasive as possible.  This came about because when I started working on AST I didn't have the main Starlink project built on my system, so I had to make a few small local patches to get anything to build.

This is an effort to clean up and formalize those patches, while trying not to break the existing in-Starlink build.  The two build systems don't fully co-exist.  Instead one starts by calling `bootstrap.local` which creates its own `configure` script.  The original `configure.ac` is mostly unmodified, but stubs are provided for the `STAR_` macros that are used.  Also had to import a copy of the messgen.c program.

Also uses a bit of a hack to only require the -starlink version of automake if `starconf` is detected on the path.

### Tests

I wanted to see if I could build the test suite via automake as well, including the Fortran tests, in the out-of-Starlink build.  This requires cherry-picking small bits of some of the Starlink libraries, which I put in `ast_tester/star_compat/`.  This was just done by trial-and-error until I had all the pieces that happen to be used by the tests.  In the process I rewrote several of the utility functions in C, partly to learn how they work and partly because I'm just more comfortable in C.  All of them are pretty straightforward, except maybe for the `psx_malloc` stuff.  This could be bug-prone since I hard-coded the size of the pointer table to 4096, which seems good enough for now.  Could change it easily enough to be dynamically resized as needed.

With these small pieces out Starlink I can successfully build and run the tests with `make check`.

I never bothered to run testhuge.f to completion (i haven't even looked at what's in it).  It's skipped by default but can be enabled by running `make check TEST_HUGE=1`.

### Misc

- Adds a minimal CI configuration that tests the standalone build on Ubuntu and macOS.  For the Mac case, although it comes with clang out of the box from Xcode it doesn't have a Fortran compiler.  I decided to give flang a try (sticking within the LLVM ecosystem) and it seems to work.

- Also found via the macOS build that recent versions of autoconf will try to automatically detect the highest C standard supported by your compiler, and will add it to the CFLAGS (e.g. `-std=gnu23`) unless you explicitly state otherwise.  This broke compiling some code (esp. in the vendored wcslib) that still used K&R style function definitions, which are no longer supported in C23.  Rather than fight autoconf on this I decided it would be worthwhile to just reformat those functions to modern style.


embray:
You can inspect a successful CI run here: https://github.com/embray/ast/actions/runs/23855735181

timj:
Thank you! I was actively pondering looking into cmake for a GitHub Action build but I'm really happy you did this.

At Rubin we are pondering trying to rewrite the Fortran tests into C to make life a bit easier for us (and maybe getting an LLM to help with that).

embray:
A cmake-based build would not be a bad idea.  I'm still partial to old school autotools, but I get comments on various other projects from people requesting a cmake build system as well, and it has its advantages (and disadvantages).

For now I was hoping to just maintain the status quo while providing more flexibility :)

embray:
Also true an LLM could likely automate rewriting most of the tests in C.  @dsberry already converted the yamlchan test in 20d3d7d63e81858f4d6e1ba8f0983c7a65b26fa0

timj:
@embray  See what you think of #38.

embray:
Wonder what to do with this once #38 is in, if any of it is worth keeping.  The changes to the autoconf build might still be worthwhile, as well as the ability to run the Fortran tests in-tree might be good, though would also be useful to integrate bits from #38 such as the header generation.

timj:
@embray Sorry if I messed you up by doing the F to C test migration and using cmake.

A starlink-less autotools build seems like it might be useful in general, especially if that means that someone can do a `make dist` for a release build and it will work without the person running that command needing Starlink software to exist.

Testing the fortran code seems like a good thing in general but is less critical now we have ctest + asan + coverage in a GitHub action now. 

I have done the K&R fixes so you can drop those from this branch when I merge (hope I can do that very soon).

Maybe we should vendor messgen and use it for the cmake build as well. I wonder if `prolat` is simple enough we could vendor that as well so that we can at least make the PDF docs. Not on this PR though.

I don't think I mind either way as to keeping this PR or not. I feel bad that you spent a lot of time getting this to work.

embray:
No worries, it's my bad for just bumbling ahead without coordinating first.  I'm glad it spurred you into doing all that other work, and there's still a lot here that can be useful I think too.  I'll wait until #38 is done and then see how to reintegrate.

ast_tester/testcmpmap.f
```diff
@@ -118,8 +118,15 @@ subroutine readobj( file, iobj, status )
       external chsource
       integer iobj, status, ch
       character file*(*)
+      integer len, chr_len
+      character srcdir*255
+      character path*255
 
-      open( 10, status='old', file=file )
+      call getenv( 'srcdir', srcdir )
```
embray: Note on this: GCC provides `getenv` as an extension so this works fine.  If for some remote reason someone couldn't use this, there is also a `psx_getenv` defined in Starlink's psx library that could easily be reimplemented in my psx_standalone.c  But it seems unnecessary.

timj: I think we use gfortran on macOS as well so it shouldn't be an issue but PSX_GETENV is the Starlink standard as you say so maybe it is better to copy that code over to psx_standalone.c since that file already exists.

embray:
@timj I rebased this on master and made several small improvements, including incorporating more of the new test fixtures (though not all of them yet).

I also make several fixes so that `make distcheck` passes, and a source tarball generated with `make dist` still works, even for the cmake build.

Remaining open question as to whether it's worth the effort to maintain two parallel testing systems.  A lot of work goes into integrating the same tests into automake as into cmake.  Though now that it's mostly done for most of the hard cases anyways, maybe just leave it for now.

I did not yet bring in some of the newer tests, including the oracle tests.  Could be left as a follow-up.

Otherwise this seems to be working well, with or without the Starlink build system.

timj:
Probably need to edit the README to describe this.

`make dist` working is good -- we need to make a new release soon since there have been lots of bug fixes. Not sure whether `cmake` is usable for making distributions.

I guess in theory we could live in a world where `make check` didn't try to do everything that `ctest` did, rather than trying to ensure that every time a new test is added it gets added to both. We could say that `ctest` is the comprehensive test suite and `make check` does a basic core test of some transforms and frameset creation to make sure that the library has built.

embray:
> We could say that ctest is the comprehensive test suite and make check does a basic core test of some transforms and frameset creation to make sure that the library has built.

I was thinking the same.  At this point I'd be a bit sad to remove the tests I already added too, but perfectly willing if we want to slim it down.  At least I finally learned how to use the mysterious TAP protocol (every time I looked at that section of the automake manual my eyes glazed over, but actually it's pretty easy).

I'll add something to the README.

I still find `make dist` to be a better way to build clean source releases than anything provided by cmake, unfortunately.

timj:
I'm okay with leaving the tests you have made in but maybe we should say that we are not required that every ctest test is also integrated into autotools. Treat ctest as the full suite going forward since it also provides the code coverage.

timj:
@embray I have rebased this ticket and made it work with the new test fixture organization. I would like to get it merged so that we have the possibility in the future to make automated releases by having git tags triggering source distribution packages to be uploaded to GitHub.

The wrinkles are:

* The version number in the build is decoupled from the git tag so we'd need to validate they match.
* The Starlink prolat tool does not exist to make the PDFs and historically the distribution tar file includes the built PDF.

I'm not sure if the prolat problem is a deal breaker. We could build the PDF separately and add it manually to the GitHub release artifacts. To make it work properly I'd probably write a little prolat replacement in python and pip install it into the CI build.

@dsberry I assume you would rather I included the PDF? I don't think including the HTML is worth doing since star2html/latex2html is very outdated now. I'd be better off porting sun_master.tex to typst.

We do need to make a release but I can do that manually for now (but I'd still like to merge this PR).

timj:
One other thing: currently the `make dist` includes the hundreds of test fixtures. That's nice for people who want to validate the build but does add a lot of volume to the distribution.

timj:
This is Claude's analysis of the situation:

# Starlink-free GNU Autotools build

Lets the autotools build run without Starlink's `starconf` and buildsupport infrastructure. `bootstrap.local` (in place of `bootstrap`) copies `star_compat/starlink_compat.m4` into `m4/` to stub the `STAR_*` macros, and `configure.ac` detects `starconf` to switch `AM_INIT_AUTOMAKE` and a `STARLINK_BUILD` conditional accordingly. Both modes still work; a Starlink build behaves as before.

62 files changed, +5566 / −314.

## Rebased onto master (`cec0f549`)

The branch predated both the CMake build and the test-fixture reorganisation. Master had moved 61 commits ahead carrying **1114 renames**, but the collision was more than renames: master moved ~1161 fixtures under `ast_tester/fixtures/` *and* changed how tests find them — CMake now stages the tree into the build directory and tests use plain relative paths — while this branch had gone the other way, deleting fixture copying in favour of a `srcdir` environment variable. The rebase keeps the branch's mechanism, re-pointed at master's paths. One commit dropped as already-upstream (41 → 40); master had kept the manifest column layouts identical, so the three TAP producers needed a new path rather than a rewrite.

The pre-rebase tip is preserved at `backup/build-system-prerebase` (`f5950748`).

## Bugs found and fixed along the way

**`F77_REAL_FUNCTION_TYPE` was wrong.** The standalone stub inferred f2c compatibility from the compiler's identity — "gfortran is f2c-compatible, so a `REAL` function returns `double`". The real Starlink macro does not guess: it compiles a C `float fred_()`, calls it from Fortran as `REAL FRED`, and reads the answer. Default gfortran is *not* in strict f2c mode, so the standalone build disagreed with both the Starlink build and CMake. AST never uses `F77_REAL_FUNCTION`, so the suite could not catch it — but `f77.h` is installed, so a downstream Fortran/C caller got a mismatched return ABI. The real probe is now ported, and CMake runs it too instead of hard-coding the value:

| | default gfortran | `FCFLAGS=-ff2c` |
| --- | --- | --- |
| autotools standalone | `float` | `double` |
| CMake | `float` | `double` |

**The documentation build had been broken for a long time.** `\htmlref` — which AST's own `addlinks` writes for *every* cross-reference — is defined only in `starstyle.4ht`, which tex4ht loads for the hypertext build. The `pdflatex` path had nothing to define it, and since TeX skips an undefined macro but still typesets its arguments, **every cross-reference was printed twice**. `\htmladdnormallink` and its footnote variant were undefined in *every* build, hypertext included. 6534 errors and a non-zero `pdflatex` exit, before:

```
                        before    after
undefined macros        6534      0
LaTeX errors            6534      0
make exit status        2         0
sun210.pdf              267 pp    266 pp   <- duplicate refs gone
sun211.pdf              274 pp    273 pp
```

**Releases shipped no LaTeX support files at all.** `starlink.cls` and friends came from `$STARLINK/share/latexsupport` via an undocumented `STARLATEXSUPPORT` environment variable that nothing set, so the `dist-hook` meant to copy them copied nothing — a release cut from this tree had documents that could not be recompiled. The four files are now vendored in `star_compat/latexsupport/`, byte-identical to upstream with provenance and SHA-256s recorded; AST's own additions live in a separate `ast.sty`, whose three macro definitions come from LaTeX2HTML 95.1's `html.sty`. `STARLATEXSUPPORT` is gone.

**`STAR_CHECK_PROGS` had its arguments reversed** in the stub, so `STAR_CHECK_PROGS(prolat, sst)` searched for a program called `sst` and `STAR_CHECK_PROGS(star2html)` set nothing. Invisible until the optional hypertext work needed `STAR2HTML`.

**Distribution completeness.** Three `EXTRA_DIST` blocks enumerated files whosereal rule is "everything of this kind", so each had to grow whenever master added one — and nothing enforced it. Both files that broke `distcheck` during this rebase (`ast_tester/testimmutable.c`, `cmake/run_simplify_noop_test.cmake`) were exactly that. Now `cmake/` ships as a tree, `ast_tester/*.{c,f,h}` ship by pattern, and `dist-check-tracked` asks git for every tracked file under `cmake/`, `ast_tester/` and `src/`, failing `make dist` immediately if one did not arrive. Verified against both a stray `.cxx` and a removed CMake helper. This also picked up ten sources that were already missing.

**Configure-generated files were being distributed.** `src/f77.h`, `src/object.h` and `ast_link_adam` all reached the tarball carrying the substitutions of whoever ran `make dist` — the Fortran ABI types, `@THREADS@`, and the `--with-external-pal` / `--without-yaml` choices respectively. Only the `.h.in` templates ship now. `src/version.h` deliberately stays: its rule is `@PREDIST@`-guarded, so an unpacked tarball cannot remake it.

**Smaller ones.** `MESSGEN` in `Makefile.am` shadowed `configure`'s own substitution, which Starlink's automake reported on every bootstrap. CMake's `TRAIL_TYPE` required a 64-bit target where neither autotools build does, so the two now emit an identical `src/f77.h`. A `.PHONY` named a target that never existed, and `CMAKE_EXTRA_FLAGS` was referenced but never defined.

## `make dist` no longer needs CMake

The CMake-from-tarball check used to run inside `make distcheck`, which meant cutting a release required both toolchains. Nothing CMake needs is generated at distribution time — unlike the autotools half-built state, it regenerates everything from `CMakeLists.txt` — so that check moved into CI, where both toolchains are present anyway.

## Verification

| Check | Result |
| --- | --- |
| `make check` (standalone autotools) | 732 / 733 pass, 1 skip (`testhuge`), 0 fail |
| `ctest` (CMake, Debug and Release) | 1251 / 1251 |
| `make distcheck` | passes, autotools only |
| CMake build of the `make dist` tarball | 1251 / 1251 |
| `make sun210.pdf sun211.pdf` | exit 0, zero LaTeX errors |
| Data-driven case counts, both suites | 192 wcsconv, 445 simplify, 20 grid |

The two `make dist` tarballs — one built with the Starlink tools, one without — differ by **11 entries out of ~1951**, every one accounted for: six documentation products and three messgen outputs on the Starlink side (both by design), and `m4/starlink_compat.m4` plus `README.md` on the standalone side. That last one is a toolchain artefact, not a repository one: automake 1.18 auto-distributes the `.md` variants of standard files and automake 1.15.1-starlink does not, so **the two tarballs cannot be byte-identical while the toolchains differ in version** — worth knowing before anyone writes a comparison into CI.

Verified on macOS 26.6.2 / Darwin 25.6.0 (arm64), Apple clang 21.0.0 and gfortran 16.1.0, automake 1.18.1 standalone and 1.15.1-starlink with autoconf 2.69 for the Starlink build.

## Notes for review

- **There are 6 `fixup!` commits** on the tip, named for the commits they belong to. They want `git rebase --autosquash --interactive master` before this merges.
- **`cmake.yaml` and `coverage.yaml` trigger changes are outside this branch's nominal scope.** All three workflows previously ran only on pushes to master and PRs targeting master, so pushing a topic branch ran nothing; they now check every branch push and every pull request, with a `concurrency` group to  cancel superseded runs. Consistency across the three seemed worth it, but say  the word and they can be split out.
- Gating merges on these still needs the check names added as required status checks in branch protection; the matrix jobs report individually.

## Deliberately not in this PR

- A Python replacement for Starlink SST's `prolat`, which `getatt` shells out  to. Until then a documentation build from a bare checkout is not possible, so  the LaTeX half being Starlink-free is the prerequisite rather than the whole  job.
- Automated releases (tag → build → upload). Everything upstream of the upload  works; the release job itself does not exist yet.
- Running every test under both build systems. 12 tests and the 152 `simplify_*_noop` idempotence checks remain CMake-only, which is intentional  for now — their sources reach the tarball, which is what matters.
- The duplicate, byte-identical `ast_test.c` at the repository root and in  `ast_tester/`.
- `TRAIL_TYPE` for non-GNU compilers: the real macro uses `long` for 64-bit  Intel and Sun Fortran and probes anything else, where both of these give every   non-GNU compiler `int`. Recorded in a comment at both sites.

embray:
@timj About the tests, previously we talked about maybe having a reasonable subset of tests as a smoke test:

> I'm okay with leaving the tests you have made in but maybe we should say that we are not required that every ctest test is also integrated into autotools. Treat ctest as the full suite going forward since it also provides the code coverage.

For the dist tarball it could include only the files needed for that subset of tests, which I could also work to formalize a bit better, such that `make distcheck` passes only on those tests, and others are skipped unless run from the git repo. I think that's reasonable.  Running the oracle tests on more different platforms would definitely be interesting too but I don't know if it needs to block install.

About the PDF doc one possibility would be to have it built and saved in git.  Yes, this is a bit of an anti-pattern but for docs I tend to make an exception, especially with latex since the output can depend a lot on what fonts, packages, and their versions are available on the system that renders it to PDF.

For some packages I've had man pages that are generated from Sphinx documentation (after all, who wants to manually write *roff?)  Rather than requiring Python + Sphinx and associated tools just to build my distribution, I keep the generated man page checked into the repository as well, and just periodically update it when necessary.  Then including it in `make dist` is trivial.  It's a compromise for practicality.

ast_tester/Makefile.am
```diff
@@ -97,14 +97,18 @@ C_TESTS = \
...
 
 # testhuge (C version) is compiled unconditionally; testhuge_f is
 # compiled when !NOFORTRAN.  Both are driven by testhuge.sh.
 check_PROGRAMS = $(C_TESTS) testhuge_c
 TESTS = $(C_TESTS)
 
+if !NOYAML
+check_PROGRAMS += testyamlchan
+TESTS += testyamlchan
+endif
```
embray, line 111: This will conflict with #67 which handles this differently but no problem for now.

makeh
```diff
@@ -259,8 +259,8 @@ unlink "/tmp/ast-$$.h";
 sub protect_copy_file {
    my $file = shift;
 
-   my $path = "$srcdir/$file";
-   $path = "$builddir/$file" if ( ! -f $path && -f "$builddir/$file" );
+   my $path = "$builddir/$file";
+   $path = "$srcdir/$file" if ! -f $path;
```
embray, line 263: Oops--yes that's better.

star_compat/starlink_compat.m4
```diff
@@ -0,0 +1,227 @@
...
+dnl
+dnl This cannot be read off the compiler's identity: it is a property of the
+dnl calling convention, and gfortran uses the f2c one only when given -ff2c.
+dnl So measure it, as the real macro does -- define a C function returning a
+dnl float, call it from Fortran as a REAL function, and see what arrives.  A
+dnl float-returning convention hands back the 1.0 that was sent; a
+dnl double-returning one reads the float's bits as a double and gets something
+dnl else, which the test detects by the result comparing equal to zero.  That
+dnl is indirect, but it is the original's test and it distinguishes the two
+dnl conventions in practice.
```
embray, line 129: Ah, this looks like something Claude wrote.  I was confused in https://github.com/Starlink/ast/pull/37#issuecomment-5483577520 if "Bugs found and fixed along the way" were things I had fixed, or things it fixed while rebasing on master...

Here there is definitely a degree of subtlety I was not aware of.  I just know enough Fortran to get as far as I did, not these kinds of nitpicks.  But I think that might explain why I was seeing some compiler warnings about this.

embray: OK I see now the full version of this is verbatim (other than the comment) out of the original starconf.m4 .  That reminds me that back when I was doing this originally I just started out with all those macros as stubs (or even missing) and tried to reverse engineer minimal stand-ins where required.  Here it's clear that the full version is more robust and makes fewer assumptions.  I'm surprised by the `R .NE. 0.0` check as opposed to just writing `R .EQ. 1.0` but I'm not sure off my head how well defined it is if a float is cast to a double/REAL.

timj:
> About the PDF doc one possibility would be to have it built and saved in git.

I wrote a new package https://github.com/Starlink/starprolog that replaces the Fortran prolat command and the AST perl getatt so I think I can now make the docs in the action.

timj:
@embray I've had Claude modify things so that `make dist` creates a distribution tarball that has only a subset of the tests. `make fulldist` will include everything. `make check` will then test what it has rather than the full suite.

I'll see if I can get the PDF build sorted out on another PR once this merges.

embray:
Oh great, I was going to maybe do something similar.

ast_tester/fixtures/DIST_MANIFEST
```diff
@@ -0,0 +1,43 @@
...
+# The header oracle corpus.  Also supplies the grid head files and the SIP
+# headers that testfitschan.c and testfitschan.f read directly.
+wcsconv/inputs/
+oracle/headers.oracle
+oracle/transform_oracle_overrides.txt
+oracle/README.md
+
+# Manifests for the families whose fixtures are not distributed.  They ship so
+# that ctest can enumerate those tests and report them as skipped rather than
+# omitting them silently; 84K buys an honest test count.
```
embray: "84K buys an honest test count" ???

ast_tester/Makefile.am
```diff
@@ -0,0 +1,717 @@
...
+	    list=`cd "$(srcdir)" && $(SHELL) ./dist-fixtures.sh \
+	        fixtures/DIST_MANIFEST` || exit 1; \
+	    tier=cutdown; \
+	fi; \
+	for rel in $$list; do \
+	    $(MKDIR_P) "$(distdir)/`dirname $$rel`"; \
+	    cp -p "$(srcdir)/$$rel" "$(distdir)/$$rel" || exit 1; \
+	done; \
+	echo $$tier > "$(distdir)/fixtures/DIST_TIER"
+	@list=`cd "$(srcdir)" && find . -maxdepth 1 -type f \
```
embray: I don't follow this part, up to the `done`.  I understand *what* it's doing but I don't understand why.  Ok for the fixture files, those have to be listed by `dist-fixtures.sh` and can't be manually enumerated.
But this seems to be automatically including *any* `.c`, `.f` etc. files (and `.awk`? where did that come from) even regardless if they're tracked in git.

Automake will already package any source files already listed as sources for tests.  I figured the sources for all the tests would be included in the tarball at least, just not necessarily all the fixtures, and that those outside the 'full' dist would be skipped.

But maybe there's something I'm missing here.

ast_tester/fixtures/DIST_MANIFEST
timj: Tweaked.

timj:
@embray  Pushed a couple more changes. If this passes are you happy to merge?

timj:
The autotools CI is a bit slow (7 minutes) but we can parallelize it a bit later.

timj:
@embray I merged this to unblock my release work now that #73 has merged. We can always tweak it later.

PR #39, opened 2026-04-13
yamlchan: Support divide transform

embray:
This adds support for the `http://stsci.edu/schemas/asdf/transform/divide-1.2.0` transform; see https://www.asdf-format.org/projects/asdf-transform-schemas/en/latest/generated/schemas/divide-1.2.0.html

This is a simple arithmetic transform that combines the outputs of two other transforms and divides them output axis-wise.

There are three other similar basic arithmetic transforms "add", "subtract", and "multiply".  I only implemented divide here because it happens to be used for the Roman WCS where it's used to implement a perspective projection step like:

```yaml
!transform/divide-1.2.0 # (x, y, z) -> (x/x, y/x, z/x)
  forward:
  - !transform/remap_axes-1.3.0  # identity: (x,y,z) -> (x,y,z)
    mapping: [0, 1, 2]
  - !transform/remap_axes-1.3.0  # (x,y,z) -> (x,x,x)
    mapping: [0, 0, 0]
    n_inputs: 3
```

I wonder if this could be done in a simpler way than I came up with, but I was trying to see how far I could get with AST's existing MathMap, and PermMaps to reorder the inputs.  Maybe could use an `IntraMap` but I didn't really dig into that yet.

Besides spherical_cartesian this was the only transform missing from yamlchan that is used in these files (not that there aren't other missing transforms).

This may have some conflicts with #35 so my plan was to get that merged first and then continue with this PR.  Or we could combine them--whichever is easier.

### TODO

* [x] Support writing the divide transform for round-tripping -- I still haven't thought about how to do that
* [x] Add tests (I have tested this works in my comparison against GWCS, but needs a stand-alone test)
* [x] Maybe support add/subtract/multiply?  This would be very easy to do with just a little refactoring, but not sure if it's worth doing here or as a follow-up

### DIFF
--- a/ast_tester/CMakeLists.txt
+++ b/ast_tester/CMakeLists.txt
@@ -365,6 +365,9 @@ endif()
 
 # --- Conditional tests ---
 if(HAVE_YAML)
+    configure_file(imaging_wcs.asdf "${CMAKE_CURRENT_BINARY_DIR}/imaging_wcs.asdf" COPYONLY)
+    configure_file(lsst_wcs.txt "${CMAKE_CURRENT_BINARY_DIR}/lsst_wcs.txt" COPYONLY)
+    configure_file(tanSipWcs.txt "${CMAKE_CURRENT_BINARY_DIR}/tanSipWcs.txt" COPYONLY)
     ast_add_test(testyamlchan)
 endif()
 
--- a/ast_tester/CMakeLists.txt
+++ b/ast_tester/CMakeLists.txt
@@ -388,6 +388,7 @@ if(HAVE_YAML)
     configure_file(imaging_wcs.asdf "${CMAKE_CURRENT_BINARY_DIR}/imaging_wcs.asdf" COPYONLY)
     configure_file(tanSipWcs.txt "${CMAKE_CURRENT_BINARY_DIR}/tanSipWcs.txt" COPYONLY)
     configure_file(lsst_wcs.txt "${CMAKE_CURRENT_BINARY_DIR}/lsst_wcs.txt" COPYONLY)
+    configure_file(rotate_seq3d_cartesian.asdf "${CMAKE_CURRENT_BINARY_DIR}/rotate_seq3d_cartesian.asdf" COPYONLY)
     ast_add_test(testyamlchan)
 endif()
 
--- a/ast_tester/CMakeLists.txt
+++ b/ast_tester/CMakeLists.txt
@@ -368,6 +368,7 @@ if(HAVE_YAML)
     configure_file(imaging_wcs.asdf "${CMAKE_CURRENT_BINARY_DIR}/imaging_wcs.asdf" COPYONLY)
     configure_file(lsst_wcs.txt "${CMAKE_CURRENT_BINARY_DIR}/lsst_wcs.txt" COPYONLY)
     configure_file(tanSipWcs.txt "${CMAKE_CURRENT_BINARY_DIR}/tanSipWcs.txt" COPYONLY)
+    configure_file(rotate_seq3d_cartesian.asdf "${CMAKE_CURRENT_BINARY_DIR}/rotate_seq3d_cartesian.asdf" COPYONLY)
     ast_add_test(testyamlchan)
 endif()
 
--- a/src/yamlchan.c
+++ b/src/yamlchan.c
@@ -474,6 +474,7 @@ static int FindAffine( int, int *, AstMapping **, int *, int * );
 static int FindDivide( int, int *, AstMapping **, int *, int * );
 static int FindRotate3d( int, int *, AstMapping **, int *, int * );
 static int FindSphericalCartesian( int, int *, AstMapping **, int *, int * );
+static void CompactMapList( int *, AstMapping **, int * );
 static int Get0I( AstKeyMap *, const char *, int, int, int * );
 static int Get1A( AstKeyMap *, const char *, int, int, AstKeyMap **, int *, int * );
 static int Get1D( AstKeyMap *, const char *, int, int, double *, int *, int * );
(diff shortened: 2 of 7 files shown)


embray:
Write support works now but it's a bit ugly.  Continuing to work on this also revealed some opportunities for refactoring to make things a little bit cleaner as the number of these "complicated" proxied transforms (that don't directly equate to existing mappings in AST) grows.

At this point it made more sense to work on top of my branch for #35 so this branch is rebased on top of it.  Should be easier to review then once #35 is merged.

embray:
Looks like I broke something while rebasing... (fixed)

dsberry:
@embray I think MathMap is the only way to go at the moment, without creating new specialised Mapping classes for +.-.*./. 

The problem with these transformations is that they have no real inverse. I see you have told MathMap to use an inverse of "p=q", "r=1.0".  There are obviously many other possible "inverses" (e.g. "p=q*q", "r=q") Does your use case actually require an inverse?  If not, it may be better simply not to define an inverse within the MathMap. This is in some ways safer as it will cause an error message to be generated if an attempt is made to use the MathMap's inverse transformation, rather than silently using a potentially inappropriate inverse.

embray:
Indeed...  According to the schema for this transform

> Invertibility: This transform is not automatically invertible.

So officially it declines to specify an inverse.  I'll try taking it out and see what happens.  I just thought it was a "reasonable" inverse if one wanted to have one as a fallback, but I agree that it's better to decline to define one, especially if user code can provide one later if needed, with their own definition.

embray:
* rebased on master
* removed inverse transform for divide
* added to and corrected history documentation

src/yamlchan.c
```diff
@@ -1033,7 +1034,7 @@ static AstMathMap *GetRefMap( AstYamlChan *this, const char *key, int *status )
 
 /* Otherwise construct and cache the appropriate MathMap. */
    } else if( !strcmp( key, REFMAP_DIVIDE_1D ) ) {
-      ref = (AstMathMap *) astMathMap( 2, 1, 1, DIVIDE_1D_FWD, 2, DIVIDE_1D_INV,
+      ref = (AstMathMap *) astMathMap( 2, 1, 0, DIVIDE_1D_FWD, 2, NULL,
```
embray: This is wrong -- I misremembered which parameter is which.  Fixing now.  Though it makes me question the fact that the tests are still passing...

embray: In fact the test is failing but is being reported as a false positive so something to fix there too.

timj:
@embray I did fix a buffer overrun in the yaml code in #40 so maybe try rebasing if you haven't already.

embray:
@timj Yep, saw that.  Just rebased.

embray:
Hmm, now ASAN is reporting a stack-buffer-overflow from testyamlchan.  It doesn't look likely to be related to anything I changed here though, though it's not inconceivable...  Consistent between macos and ubuntu too which is interesting.  Will see if I can reproduce...

embray:
I get this on the master branch as well.

timj:
You mean you get the ASAN error on master branch without any other changes? Or on master branch with enhanced yaml test file?

A quick glance indicates that dimw/dimd should be 2 element arrays and not scalars. Might be able to look after work.

embray:
I'm getting it on master without any other changes.  That's what's strange.  I'm investigating.

embray:
@timj Yep, right call.  That fixes it, plus a couple other typos.  Now looking at a different one that comes up...

embray:
1556f3f677452ab5356ad9e0306fb6d45d04b1ab added a few other miscellaneous fixes from my development branch that I forgot about.  Easier to include on this PR since there would be merge conflicts otherwise.

embray:
Rebased again.

timj:
Do you have strategy for testing YamlChan more fully? I noticed in the coverage report that coverage is poor in general (I haven't checked the coverage report from this branch). Could we get away with reading some ASDF examples and then serializing them back to ASDF and comparing?

timj:
I realize I still need to tidy up the ast_tester directory to put test fixtures in sub directories consistently.

ast_tester/CMakeLists.txt
```diff
@@ -382,6 +382,10 @@ endif()
 
 # --- Conditional tests ---
 if(HAVE_YAML)
+    configure_file(imaging_wcs.asdf "${CMAKE_CURRENT_BINARY_DIR}/imaging_wcs.asdf" COPYONLY)
+    configure_file(lsst_wcs.txt "${CMAKE_CURRENT_BINARY_DIR}/lsst_wcs.txt" COPYONLY)
+    configure_file(tanSipWcs.txt "${CMAKE_CURRENT_BINARY_DIR}/tanSipWcs.txt" COPYONLY)
+    configure_file(rotate_seq3d_cartesian.asdf "${CMAKE_CURRENT_BINARY_DIR}/rotate_seq3d_cartesian.asdf" COPYONLY)
```
embray, line 391: @timj Part of the existing dismal low coverage reported on this file is because this was missing, combined with the missing return statement I added at https://github.com/Starlink/ast/pull/39/changes#diff-2ee0d9ab522f35baa976846063afdc3c96b610076a70152c26ba48c3c542b7f3R56

So most/all the yamlchan tests were failing silently.  

With these fixes the line coverage is shown at 66%

Image

Still plenty of room for improvement.  A few more tests of individual transforms in the same spirit of the tests I've already added would probably make a difference.  Undecided whether to make a single fixture file with a bunch of different transforms in it, or several small ones focused on individual transforms like `rotate_seq3d_cartesian.asdf`.  Testing more of the projection transforms and frame types is also in order.

embray:
Is it OK if I go ahead and merge this?  It shouldn't impact much, but I have a few other in-progress works that depend on it (particularly the `WriteProxy` refactoring).

timj:
Okay with me to merge.

PR #41, opened 2026-04-21
Add some more test coverage

timj:
* Test astResample (use smaller variant of testhuge)
* Add some more example FITS headers

PR #42, opened 2026-04-22
fitschan: Prevent failure when reading FITS-CLASS headers

dsberry:
FITS-CLASS headers are only defined for the primary axis descriptions. Attempting to read them for alternate axis descriptions was causing an error. This PR contains a fix for this bug plus an extra wcs conversion test.

### DIFF
--- a/ast_tester/ast_tester
+++ b/ast_tester/ast_tester
@@ -177,7 +177,8 @@ foreach n ( "timj ast fits-wcs cdmatrix=1" \
             "sip head fits-wcs cdmatrix=1,sipreplace=0" \
             "sip2 head fits-wcs cdmatrix=1,sipreplace=0" \
             "lsst ast fits-wcs cdmatrix=1" \
-            "longslit fits-pc fits-wcs cdmatrix=1" )
+            "longslit fits-pc fits-wcs cdmatrix=1" \
+            "jan fits-class ast" )
 
   set a = `echo $n`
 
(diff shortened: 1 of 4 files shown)


timj:
The incredible thing is that last night in #41 I was noticing this exact problem when I tried to expand the test coverage of fitschan.c by using some CLASS headers. It took Claude half an hour of thought to come up with a slightly different fix that you can see in https://github.com/Starlink/ast/pull/41/changes/91bad7e4e0aba39db9ba79e5fd8b073588610c30

```c
if( encoding == FITSCLASS_ENCODING && axlat >= 0 && axlon >= 0 ) 
```

I'll drop that fix and rebase on top of this one.

dsberry:
> The incredible thing is that last night in #41 I was noticing this exact problem when I tried to expand the test coverage of fitschan.c by using some CLASS headers. It took Claude half an hour of thought to come up with a slightly different fix that you can see in [91bad7e]

That makes me feel a bit better - I think it only took me about 10 minutes!

(https://github.com/Starlink/ast/commit/91bad7e4e0aba39db9ba79e5fd8b073588610c30)
> 
> ```c
> if( encoding == FITSCLASS_ENCODING && axlat >= 0 && axlon >= 0 ) 
> ```
> 
> I'll drop that fix and rebase on top of this one.

I think both fixes will be functionally equivalent in practice.

PR #43, opened 2026-04-24
ast_tester: add new utility function to help configure plplot

embray:
Side quest I went on while working on test coverage support.

I found that the plotting tests were hanging for me -- even though I had PLplot installed, I did not have the cairo driver for it, and the tests that generate PDFs all default to the pdfcairo driver.

PLplot's default behavior in this case, a bit irritatingly, is to block on a stdin prompt requesting the user to select a driver.

After digging around I found some public functions in the C API that allow probing for supported devices.  At first I just put this support code in the tests, but there was some duplication in both testplot3d.c and testplotter.c, so I thought maybe this could make a useful addition to the public API.

After getting that working, it turned out that some of the plots fail with PLplot's non-cairo PDF backend, libharu.  Understanding why that is at this point is beyond me though, so I added a configure-time check for this and marked the affected tests as expected failures in this case.

### DIFF
--- a/ast_tester/CMakeLists.txt
+++ b/ast_tester/CMakeLists.txt
@@ -3,6 +3,7 @@
 # These are the existing C test programs, plus Fortran tests that have
 # been converted to C.  All tests use CTest and do not require any
 # Starlink infrastructure.
+include(CheckCSourceRuns)
 
 # Ensure test executables find the locally-built libast before any
 # system-installed version (e.g. from a conda environment whose lib
@@ -465,8 +466,37 @@ foreach(_entry IN LISTS _grid_tests)
 endforeach()
 
 if(PLPLOT_FOUND)
+    # --- Determine if pdfcairo is available ---
+    # If not, plplot falls back on the "pdf" device which uses libHaru, which
+    # fails with some of the pdf tests; if the pdfcairo device is not availble
+    # then we just mark these tests as expected failures.
+
+    check_c_source_runs("
+    #include <string.h>
+    #include <plplot.h>
+    #include \"ast.h\"
+
+    int main(void) {
+        const char *dev = astPlSetupDevice(\"test.pdf\");
+        if (dev && strcmp(dev, \"pdfcairo\") == 0) {
+            return 0;
+        } else {
+            return 1;
+        }
+    }
+    " HAVE_PLPLOT_PDFCAIRO)
+
+    # Plotter tests that are known to fail without pdfcairo
+    set(PLOTTER_XFAIL_NO_PDFCAIRO
+        car4
+        car6
+        specflux
+        timeplot
+    )
+
     configure_file(lsst.fits-wcs "${CMAKE_CURRENT_BINARY_DIR}/lsst.fits-wcs" COPYONLY)
     ast_add_test(testplotter
+        SOURCES testplotter.c plplotutil.c
         NO_DEFAULT_GRF_LIBS
         EXTRA_LIBS $<LINK_LIBRARY:WHOLE_ARCHIVE,ast_grf_plplot> $<LINK_LIBRARY:WHOLE_ARCHIVE,ast_grf3d_plplot> ${PLPLOT_LIBRARIES}
         TEST_ARGS "lsst.fits-wcs" " " " " "testplotter.ps"
@@ -476,6 +506,7 @@ if(PLPLOT_FOUND)
 
     configure_file(plot3d-test1.ast "${CMAKE_CURRENT_BINARY_DIR}/plot3d-test1.ast" COPYONLY)
     ast_add_test(testplot3d
+        SOURCES testplot3d.c plplotutil.c
         NO_DEFAULT_GRF_LIBS
         EXTRA_LIBS $<LINK_LIBRARY:WHOLE_ARCHIVE,ast_grf_plplot> $<LINK_LIBRARY:WHOLE_ARCHIVE,ast_grf3d_plplot> ${PLPLOT_LIBRARIES}
         TEST_ARGS "testplot3d.pdf"
@@ -520,9 +551,20 @@ if(PLPLOT_FOUND)
                 "-DBOX=${_box_arg}"
                 -P "${CMAKE_SOURCE_DIR}/cmake/run_plotter_test.cmake"
             WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}")
+
         if(AST_ENABLE_SANITIZERS)
             set_tests_properties(plotter_${P_NAME} PROPERTIES
                 ENVIRONMENT "ASAN_OPTIONS=detect_leaks=0")
+        endif()
+
+        # Mark test as expected failure if it's one known not to work
+        # without pdfcairo
+        if(NOT HAVE_PLPLOT_PDFCAIRO)
+            list(FIND PLOTTER_XFAIL_NO_PDFCAIRO "${P_NAME}" _idx)
+            if(NOT _idx EQUAL -1)
+                set_tests_properties(plotter_${P_NAME} PROPERTIES
+                    WILL_FAIL TRUE)
+            endif()
         endif()
     endfunction()
 
@@ -546,6 +588,5 @@ if(PLPLOT_FOUND)
         endif()
         add_plotter_test(${_args})
     endforeach()
-
 endif()
 
--- a/ast_tester/CMakeLists.txt
+++ b/ast_tester/CMakeLists.txt
@@ -471,20 +471,31 @@ if(PLPLOT_FOUND)
     # fails with some of the pdf tests; if the pdfcairo device is not availble
     # then we just mark these tests as expected failures.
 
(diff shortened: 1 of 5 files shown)


ast_tester/CMakeLists.txt
```diff
@@ -461,6 +462,34 @@ foreach(_entry IN LISTS _grid_tests)
...
+
+    int main(void) {
+        const char *dev = astPlSetupDevice(\"test.pdf\");
+        if (dev && strcmp(dev, \"pdfcairo\") == 0) {
+            return 0;
+        } else {
+            return 1;
+        }
+    }
+    " HAVE_PLPLOT_PDFCAIRO)
```
embray, line 495: Obviously this was never going to work because it requires ast to be built in the first place.  I'll think of something else...

embray:
This is ready from my side; I think the coverage complaint is a bit superfluous in this case.

ast_tester/CMakeLists.txt
```diff
@@ -461,6 +462,45 @@ foreach(_entry IN LISTS _grid_tests)
 endforeach()
 
 if(PLPLOT_FOUND)
+    # --- Determine if pdfcairo is available ---
+    # If not, plplot falls back on the "pdf" device which uses libHaru, which
+    # fails with some of the pdf tests; if the pdfcairo device is not availble
+    # then we just mark these tests as expected failures.include(CheckCSourceRuns)
```
timj: Looks like the end of this sentence is a paste error.

What failures happen with the different pdf device?

src/grf_plplot.c
```diff
@@ -98,9 +110,176 @@ static char *format_text(const char *in_text) {
...
+/* ========================================================= */
+
+/* Initialize global array of available plplot devices */
+static void astPlInitDeviceList( void ) {
+   const char **menustr;
+
+   if (pl_devnames)
+      return;
+
+   menustr = calloc(pl_ndev, sizeof(*menustr));
```
timj: Probably better to use the ast memory APIs for consistency with the rest of the file.

src/grf_plplot.c
```diff
@@ -98,9 +110,176 @@ static char *format_text(const char *in_text) {
...
+         return "null";
+      }
+   }
+
+   ext = strrchr( output, '.' );
+   dev = astPlChooseDeviceForExtension( ext );
+
+   if ( !astPlDeviceAvailable( dev ) ) {
+      astError( AST__GRFER, "astPlSetupDevice: plplot output device %s for "
+                "%s files not available; falling back on null device\n", ext,
```
timj:
```suggestion
                "%s files not available; falling back on null device\n", dev,
```
?

src/grf_plplot.c
```diff
@@ -98,9 +110,176 @@ static char *format_text(const char *in_text) {
...
+*        - If all else fails outputs to the "null" device.
+
+*  Returned Value:
+*     The name of the device selected.
+*/
+const char *astPlSetupDevice( const char *output ) {
+   const char *ext;
+   const char *dev;
+
+   if( strcmp(output, "aqt") == 0 || strcmp(output, "xwin") == 0 ||
```
timj: Do we have to check that `output` is not NULL first?

src/grf_plplot.c
```diff
@@ -98,9 +110,176 @@ static char *format_text(const char *in_text) {
...
+       strcmp(output, "xcairo") == 0 || strcmp(output, "qtwidget") == 0 ) {
+
+      if (astPlDeviceAvailable( output )) {
+         c_plsdev( output );
+         return output;
+      } else {
+         astError( AST__GRFER, "astPlSetupDevice: selected plplot output "
```
timj: Doesn't astError set the global AST status to bad? So all the calls in the test code are going to be skipped even though you are returning a default device? I don't think the return value is used though. Should the test code return with error immediately if AST global status is set?

CMakeLists.txt
```diff
@@ -416,6 +416,7 @@ set(AST_H_FILES
     src/frameset.h
     src/plot.h
     src/plot3d.h
+    src/grf_plplot.h
```
timj: Does this lead to all the plplot specific stuff ending up in ast.h as public interface? That seems wrong.

src/grf_plplot.c
```diff
@@ -98,9 +110,176 @@ static char *format_text(const char *in_text) {
...
+   pl_devnames = calloc(pl_ndev, sizeof(*pl_devnames));
+
+   if ( !pl_devnames ) {
+      free( menustr );
+      return;
+   }
+
+   /* Not-well advertised but available API for probing supported devices */
+   plgDevs(&menustr, &pl_devnames, &pl_ndev);
+   free( menustr );
```
timj: In theory I think we need to free pl_devnames.

timj (approved):
Thanks for this. Remember to add a HISTORY prologue entry.

My main thing I'm wondering is whether this stays completely private for build testing vs having plplot-specific interfaces turning up in ast.h (which seems wrong).

ast_tester/CMakeLists.txt
embray: Yep, looks like i must have fat-fingered a 'p', will fix.

The failures all look something like this:

```
122/137 Testing: plotter_car4
122/137 Test: plotter_car4
Command: "/usr/bin/cmake" "-DTESTPLOTTER=/home/embray/src/starlink/ast/build_cov/ast_tester/testplotter" "-DHEAD_FILE=car4.head
" "-DATTR=format(1)=gd,format(2)=gd,Grid=1,tickall=0,width(axes)=3" "-DFATTR=carlin=1" "-DOUT_FILE=plotter_car4.pdf" "-DBOX=0 0 951 1851" "-P" "/home/embray/src/starlink/ast/cmake/run_plotter_test.cmake"
Directory: /home/embray/src/starlink/ast/build_cov/ast_tester
"plotter_car4" start time: Apr 24 15:13 CEST
Output:
----------------------------------------------------------
ERROR in haru library
ERROR: error_no=1050, detail_no=0
CMake Error at /home/embray/src/starlink/ast/cmake/run_plotter_test.cmake:35 (message):
  testplotter exited with code 1 (cmd:
  /home/embray/src/starlink/ast/build_cov/ast_tester/testplotter;car4.head;format(1)=gd,format(2)=gd,Grid=1,tickall=0,width(axes)=3;carlin=1;plotter_car4.pdf;0;0;951;1851)
```

Of course, I wondered if that could/should just be fixed, but getting into the weeds of how the PDF generation works is beyond what I can really be spending my time on unfortunately so I thought better to just skip these as known failures for now.  Obviously if there are any better ideas for fixing it that would be better.

src/grf_plplot.c
embray: But where?  Or maybe just don't bother caching it globally.  I looked around but found that AST doesn't (currently) use destructor functions anywhere.

src/grf_plplot.c
embray: Yes, thanks.  I think an earlier version of this had an early escape hatch for `output == NULL` but I forgot it when moving stuff around.

src/grf_plplot.c
embray: You're right, it probably doesn't make any sense to make this an `astError`. 

For some background, originally I just had this code directly in the test code only.  Then at some point I thought it might be useful to have in the main library code and moved it here.  Originally this was just a printf to stderr.  But it was also late Friday afternoon and I think my brain was already half-gone ;)

I think at most this was intended to be a warning message, but it turns out AST doesn't really have a logging facility for stuff like this...

src/grf_plplot.c
embray: should be `dev, ext`

CMakeLists.txt
embray: Only the new `const char *astPlSetupDevice( const char * );` function declaration.

Of course, if you're on the fence about whether or not AST should export any plplot-specific functionality in favor of keeping the graphics backend totally transparent, I could take this out.

Like I wrote above originally this was just written as a utility to use in the tests since the existing device selection code was duplicated in a couple places, and flaky.  I thought it might make a useful addition though for when one happens to know they're using plplot, since:

1) The actual plplot API for this is not very good -- actually quite hostile for library authors
2) It is encoding what seem to be best practices in terms of selecting the best backend devices for plplot for different output formats

timj: I would prefer we not add backend-specific APIs into the public ast.h. A public  `star/ast_plplot.h` would be better I think. It's not like PLplot is being developed -- it looks like a dead project to me but I needed something to test with that wasn't pgplot. Then I ended up also making a SVG backend that didn't need any dependencies.

embray:
Ok, I'll just move this back to being a utility used for the tests, which is where it started out.  It's good to have because it still leads to fewer surprises (tests hanging)

embray:
I removed the PLplot device setup stuff out of the main library and just put it back into a utility module to use in the tests.

timj (approved):
Looks great. I can merge if you are ready

embray:
Yep!

Issue #46, opened 2026-05-08
Possible sph_fwd_inv_unitradius_cancel fixture discrepancy


timj:
@dsberry can you confirm that this simplification test looks right to you? The thing that jumps out at me is that the input .map has 3 inputs but the simplified form has 2 inputs. How can you lose inputs on simplification? It looks likes it's a SphMap 3 in to 2 out followed by inverted SphMap going from 2 back to 3 so UnitMap(3) simplification seems right and the actual simplification of UnitMap(2) seems wrong.

## Fixture files

- **Input:** `ast_tester/simplify_fixtures/sph_fwd_inv_unitradius_cancel.map`
- **C reference:** `ast_tester/simplify_fixtures/sph_fwd_inv_unitradius_cancel.simp`

## Input mapping (native format)

```
Begin CmpMap
    Nin = 3
 IsA Mapping
    InvB = 1
    MapA =
       Begin SphMap
          Nin = 3
          Nout = 2
       IsA Mapping
          UntRd = 1
       End SphMap
    MapB =
       Begin SphMap
          Nin = 3
          Nout = 2
          Invert = 1
       IsA Mapping
          UntRd = 1
       End SphMap
End CmpMap
```

## C reference simplification

```
Begin UnitMap
    Nin = 2
    IsSimp = 1
 IsA Mapping
End UnitMap
```

timj:
Fixed in #47.

PR #47, opened 2026-05-09
sphmap: Fix simplification bug

dsberry:
Fix bug in MapMerge - UnitMap that replaces back to back SphMaps was not taking account of the direction of the two SphMaps. It always assumed the first was forward and the second inverse, and thus simplified to a 2-in 2-out UnitMap. If they are the other way round, then the UnitMap should have 3 ins and outs.


timj:
Thanks. Tests are failing because the .simp file has not been updated yet.

dsberry:
Presumably that's ready to merge now?

timj:
Closes #46

PR #48, opened 2026-05-09
ast_tester: Only create plots for 2-dimensional headers

dsberry:
Previously, plots were attempted for all *.head files in ast_tester. But a lot of these files are now 1-dimensional and so could not be plotted. This change specifies an explicit list of 2-dimensional headers to plot.


timj:
Ah yes. Because I created hundreds of example FITS headers for code coverage.

PR #51, opened 2026-06-03
Experimental optional vectorization for SphMap, PolyMap and MatrixMap

embray:
## Background

A couple weeks ago I was doing some benchmarking of AST (wrapped in my libasdf-gwcs library) against the Python GWCS package.  I found, to no surprise, that for many cases AST far outperformed Python (on a single thread).  However, at about N=100000 points the Python code started to outperform slightly.

I concluded that this was due to NumPy's built-in SIMD support for many of its ufuncs (and potentially BLAS as well, though I don't think that was a major contributor since there are only some small matrix multiplications performed).  On my machine N=100000 happened to be where the Python overhead is amortized by the SIMD enhancements.

I found that the major contributors in my case were PolyMap, MatrixMap, and SphMap, with the biggest potential win coming from SphMap.  So I set about adding some bare minimum compiler (GCC) generated vectorization in targeted areas to see if we could get any quick wins, starting with SphMap, and indeed it did make a big difference.

I've added a little benchmark program adapted from my libasdf-gwcs benchmark.  I then went on to see if I could get any easy wins with MatrixMap and PolyMap.  These were a little trickier to get right, especially PolyMap.  The key observation in all these cases is that the main outer loop of each `Transform` method is over the input points.  In a way this makes sense for cache efficiency, but ignoring that for the first pass I trudged ahead, and did close the gap with Python up to a point--eventually at even larger N the performance was being dominated by DRAM moves.

At least for MatrixMap and PolyMap I added a utility function to return the system's L2 cache size (should work for any current x86-64, YMMV on other platforms), and add a separate outer loop to chunk the points into sets small enough to fit in cache (assuming not huge numbers of polynomial coeffs or matrix elements, other overhead), and this proved a big win--AST at parity with or better than NumPy, with near consistent throughput and memory usage even for large N.

## End-to-end impact (libasdf-gwcs vs Python GWCS)

The real motivation, and the most representative measurement, is full Roman WCS evaluation through libasdf-gwcs (which drives AST's `astTran2` over a complete WCS pipeline on real Roman L2 calibration files).  The plots below show single-thread throughput vs Python GWCS across N, before and after this change.

Before:

throughput

After, with AST compiled with SIMD features:

libasdf-gwcs vs Python GWCS -- after AST SIMD

The above was just from one run of the benchmark but YMMV. In effect this closes the gap with NumPy, and I had some runs depending on the conditions that outperformed it entirely.

The per-transform microbenchmarks further down isolate where the gains come from; this end-to-end view is what actually matters for a WCS pipeline, and is also where the L2 chunking pays off most (each pipeline stage's output stays cache-resident as the next stage's input, rather than being re-streamed from DRAM).

## What's in this PR

- **Opt-in build flag.**  A new `AST_ENABLE_SIMD` option (CMake and autotools), **OFF by default**.  When enabled, SIMD-friendly compiler flags (`-fopenmp-simd -ffast-math -fno-associative-math -march=x86-64-v3`) are applied *only* to `sphmap.c`, `polymap.c` and `matrixmap.c`, so the floating-point semantics of the rest of the library are unchanged.  The flags are only accepted on a GCC/glibc toolchain that supports them; otherwise the option is a no-op.

- **`TransformLoop` vtable slot.**  Each of SphMap, PolyMap and MatrixMap gains a per-class `TransformLoop` slot.  `Transform` is now a thin wrapper (validation, invert handling, point extraction) that dispatches to either `TransformLoopScalar` or `TransformLoopSIMD`.  This keeps the scalar and vectorized inner loops cleanly separated instead of `#ifdef`-ing through the body of `Transform`.

- **Runtime `UseSIMD` attribute.**  A per-instance integer attribute on each of those classes (default 1 when built with SIMD support, 0 otherwise) lets a caller force the scalar path at runtime -- useful for bit-exact reproducibility or debugging without rebuilding.  Setting `UseSIMD=1` on a library built without SIMD reports `AST__ATSER`.  The attribute is deliberately not serialized (it's just a runtime tuning knob, not part of the mapping description).  I debated making this some kind of global setting but opted to make it specific to the supported transforms for now.

  - Between `TransformLoop` and `UseSIMD` there's a bunch of duplicative code.  I didn't want to do any more refactoring for now before waiting to see what you have to say about this, but it seems to merit a base class or even just moved into the base `Mapping` class.

- **`astCPUCacheSize(level)`** in `memory.c`: returns the L1D/L2/L3 cache size via CPUID (x86-64), with a `sysconf` fallback and hard-coded defaults.  PolyMap and MatrixMap use the L2 size to pick a chunk size.

- **The vectorized kernels** themselves (the substance):
  - *SphMap* -- inline `atan2`/`sqrt` (forward) and `sin`/`cos` (inverse) under `#pragma omp simd`, which GCC routes through glibc `libmvec`; a scalar fixup pass handles poles and bad values.
  - *PolyMap* -- invert the nested evaluation loops so the innermost loop is over points (terms become independent and vectorize); process points in L2-sized chunks so the per-power work buffers stay cache-resident.  Falls back to scalar for small N (where the per-call buffer setup dominates) and for ChebyMap (which overrides `PolyPowers`).  This could easily be extended to ChebyMap too but I skipped it for now since my usecase doesn't use any Chebychev polynomials, at least for now.
  - *MatrixMap* -- full matrix: point loop innermost, `VFMADD` via `#pragma omp simd`, L2 chunking, and a bad-value fixup that runs only when a vectorized scan finds a bad input; diagonal matrix: a branchless compare/blend.

- **`perf/bench_simd.c`** -- a standalone benchmark (details below).

## Benchmark

`perf/bench_simd` runs each transform twice in a single process -- scalar (`UseSIMD=0`) and SIMD (`UseSIMD=1`) via the runtime attribute -- over an N sweep, and prints a scalar-vs-SIMD summary (plain text, or `--markdown`).

A note on methodology, since it matters for reading the small-N numbers: each measurement times a *calibrated batch* of repeated `astTranP` calls (the batch grows until it runs for at least 20 ms) and reports the amortized per-call time.  At small N a single call is dominated by call overhead and clock resolution rather than throughput, so batching is what makes those rows meaningful -- it removes measurement *error*, not the per-call cost itself.  The input arrays stay hot across a batch, so small/medium N measures compute-bound throughput and large N (which exceeds the last-level cache) measures bandwidth-bound throughput.  Reported figures are the median over the reps (`-r`, default 5).

**Hardware/build for the numbers below:** Intel i7-7820HQ (Kaby Lake, AVX2), L2 = 1 MiB/core, L3 = 8 MiB; GCC 13.3.0 (i.e. just my laptop); `-DCMAKE_BUILD_TYPE=RelWithDebInfo -DAST_ENABLE_SIMD=ON`.

### Headline

| Transform | what it exercises | peak SIMD speedup | large-N (16.7M) |
|-----------|-------------------|------------------:|----------------:|
| `sphmap_fwd` | (x,y,z)->(lon,lat), `atan2`/`sqrt` | \~8.6x (N\~65K) | 8.4x |
| `sphmap_inv` | (lon,lat)->(x,y,z), `sin`/`cos` | \~6.9x (N\~16K) | 6.3x |
| `poly5_fwd` | degree-5 2-D PolyMap (SIP-like) | \~3.1x (N\~1K) | 2.9x |
| `poly1_fwd` | degree-1 2-D PolyMap (linear) | \~2.7x (N\~1K) | 2.4x |
| `matfull2_fwd` | 2x2 full MatrixMap (rotation) | \~4.5x (N\~4K) | 2.8x |
| `matdiag2_fwd` | 2x2 diagonal MatrixMap (scale) | \~1.8x (N\~4K) | 1.2x |

A few things worth calling out:

- **SphMap is the big win**, as expected -- almost all of its cost is transcendental math, and `libmvec` vectorizes that directly.
- **PolyMap** sustains ~2.5-3x across the useful range.
- **MatrixMap full** went from a modest gain to ~2.8-4.5x after restructuring the kernel to write each output once and skip the bad-value fixup on clean data.
- **MatrixMap diagonal** stays modest (~1.2x at large N) because it is a single-pass streaming kernel (one read, one multiply, one write per element, no data reuse).  At large N it is simply DRAM-bandwidth-bound, and chunking cannot reduce its mandatory traffic -- this is expected, not a regression.
- Below N~16-32 the SIMD path is at best a wash; PolyMap explicitly falls back to scalar there, and the others come out ~1.0x.

### Full sweep

<details>
<summary>scalar vs SIMD, Mpx/s, N = 1 ... 16.7M (median of 9 reps)</summary>

| Transform | N | Scalar Mpx/s | SIMD Mpx/s | Speedup |
|-----------|--:|-------------:|-----------:|--------:|
| sphmap_fwd | 1 | 1.1 | 1.0 | 0.90x |
| sphmap_fwd | 4 | 3.5 | 3.4 | 0.97x |
| sphmap_fwd | 8 | 6.3 | 7.5 | 1.20x |
| sphmap_fwd | 16 | 9.3 | 16.5 | 1.77x |
| sphmap_fwd | 32 | 11.5 | 27.0 | 2.35x |
| sphmap_fwd | 64 | 14.2 | 42.8 | 3.01x |
| sphmap_fwd | 128 | 15.1 | 58.9 | 3.90x |
| sphmap_fwd | 256 | 14.9 | 83.3 | 5.60x |
| sphmap_fwd | 512 | 13.7 | 90.7 | 6.62x |
| sphmap_fwd | 1024 | 13.3 | 100.1 | 7.51x |
| sphmap_fwd | 4096 | 12.6 | 103.9 | 8.26x |
| sphmap_fwd | 16384 | 13.3 | 113.5 | 8.55x |
| sphmap_fwd | 65536 | 12.8 | 109.6 | 8.58x |
| sphmap_fwd | 262144 | 12.7 | 109.9 | 8.64x |
| sphmap_fwd | 1048576 | 12.9 | 106.5 | 8.25x |
| sphmap_fwd | 4194304 | 12.8 | 107.2 | 8.40x |
| sphmap_fwd | 16711744 | 12.4 | 104.1 | 8.41x |
| sphmap_inv | 1 | 1.1 | 1.1 | 1.02x |
| sphmap_inv | 4 | 4.0 | 4.1 | 1.01x |
| sphmap_inv | 8 | 7.0 | 8.9 | 1.28x |
| sphmap_inv | 16 | 10.5 | 16.4 | 1.56x |
| sphmap_inv | 32 | 14.7 | 28.5 | 1.94x |
| sphmap_inv | 64 | 18.2 | 43.2 | 2.38x |
| sphmap_inv | 128 | 20.8 | 65.7 | 3.16x |
| sphmap_inv | 256 | 21.6 | 87.2 | 4.03x |
| sphmap_inv | 512 | 21.9 | 101.6 | 4.63x |
| sphmap_inv | 1024 | 18.0 | 107.3 | 5.95x |
| sphmap_inv | 4096 | 17.1 | 114.8 | 6.70x |
| sphmap_inv | 16384 | 17.3 | 119.2 | 6.89x |
| sphmap_inv | 65536 | 18.0 | 119.0 | 6.62x |
| sphmap_inv | 262144 | 17.5 | 111.2 | 6.36x |
| sphmap_inv | 1048576 | 17.6 | 110.9 | 6.30x |
| sphmap_inv | 4194304 | 16.4 | 110.0 | 6.69x |
| sphmap_inv | 16711744 | 16.9 | 106.0 | 6.26x |
| poly5_fwd | 1 | 0.9 | 0.9 | 0.96x |
| poly5_fwd | 4 | 3.1 | 3.3 | 1.04x |
| poly5_fwd | 8 | 5.5 | 5.6 | 1.02x |
| poly5_fwd | 16 | 9.0 | 9.1 | 1.01x |
| poly5_fwd | 32 | 12.3 | 15.5 | 1.26x |
| poly5_fwd | 64 | 14.4 | 26.3 | 1.83x |
| poly5_fwd | 128 | 16.6 | 41.8 | 2.51x |
| poly5_fwd | 256 | 19.5 | 55.5 | 2.84x |
| poly5_fwd | 512 | 19.8 | 61.6 | 3.11x |
| poly5_fwd | 1024 | 20.6 | 64.7 | 3.14x |
| poly5_fwd | 4096 | 20.9 | 62.1 | 2.97x |
| poly5_fwd | 16384 | 21.7 | 62.4 | 2.88x |
| poly5_fwd | 65536 | 22.1 | 62.4 | 2.82x |
| poly5_fwd | 262144 | 22.4 | 60.5 | 2.70x |
| poly5_fwd | 1048576 | 20.9 | 59.8 | 2.86x |
| poly5_fwd | 4194304 | 21.3 | 59.9 | 2.81x |
| poly5_fwd | 16711744 | 20.6 | 59.6 | 2.89x |
| poly1_fwd | 1 | 0.9 | 1.0 | 1.03x |
| poly1_fwd | 4 | 3.6 | 3.5 | 0.97x |
| poly1_fwd | 8 | 6.3 | 6.2 | 0.98x |
| poly1_fwd | 16 | 10.9 | 10.2 | 0.94x |
| poly1_fwd | 32 | 15.9 | 18.8 | 1.18x |
| poly1_fwd | 64 | 23.2 | 33.9 | 1.46x |
| poly1_fwd | 128 | 28.0 | 53.5 | 1.91x |
| poly1_fwd | 256 | 31.9 | 72.7 | 2.28x |
| poly1_fwd | 512 | 34.3 | 88.8 | 2.59x |
| poly1_fwd | 1024 | 37.2 | 101.9 | 2.74x |
| poly1_fwd | 4096 | 37.7 | 99.2 | 2.63x |
| poly1_fwd | 16384 | 39.3 | 97.8 | 2.49x |
| poly1_fwd | 65536 | 39.9 | 99.4 | 2.49x |
| poly1_fwd | 262144 | 38.9 | 95.2 | 2.45x |
| poly1_fwd | 1048576 | 38.8 | 89.9 | 2.31x |
| poly1_fwd | 4194304 | 36.3 | 93.1 | 2.56x |
| poly1_fwd | 16711744 | 38.5 | 93.2 | 2.42x |
| matdiag2_fwd | 1 | 1.1 | 1.2 | 1.11x |
| matdiag2_fwd | 4 | 4.2 | 4.5 | 1.07x |
| matdiag2_fwd | 8 | 9.0 | 8.7 | 0.97x |
| matdiag2_fwd | 16 | 15.9 | 17.0 | 1.07x |
| matdiag2_fwd | 32 | 36.2 | 33.6 | 0.93x |
| matdiag2_fwd | 64 | 67.6 | 69.4 | 1.03x |
| matdiag2_fwd | 128 | 125.6 | 130.2 | 1.04x |
| matdiag2_fwd | 256 | 215.3 | 252.8 | 1.17x |
| matdiag2_fwd | 512 | 335.3 | 412.4 | 1.23x |
| matdiag2_fwd | 1024 | 457.4 | 679.5 | 1.49x |
| matdiag2_fwd | 4096 | 622.1 | 1103.5 | 1.77x |
| matdiag2_fwd | 16384 | 694.0 | 1007.3 | 1.45x |
| matdiag2_fwd | 65536 | 660.9 | 1063.3 | 1.61x |
| matdiag2_fwd | 262144 | 587.7 | 736.7 | 1.25x |
| matdiag2_fwd | 1048576 | 441.7 | 543.6 | 1.23x |
| matdiag2_fwd | 4194304 | 437.1 | 505.2 | 1.16x |
| matdiag2_fwd | 16711744 | 438.1 | 508.4 | 1.16x |
| matfull2_fwd | 1 | 1.2 | 1.0 | 0.83x |
| matfull2_fwd | 4 | 4.7 | 4.4 | 0.94x |
| matfull2_fwd | 8 | 8.4 | 8.4 | 0.99x |
| matfull2_fwd | 16 | 14.5 | 16.9 | 1.16x |
| matfull2_fwd | 32 | 28.8 | 35.1 | 1.22x |
| matfull2_fwd | 64 | 46.7 | 63.8 | 1.37x |
| matfull2_fwd | 128 | 64.1 | 122.7 | 1.91x |
| matfull2_fwd | 256 | 83.3 | 216.8 | 2.60x |
| matfull2_fwd | 512 | 100.9 | 338.4 | 3.35x |
| matfull2_fwd | 1024 | 108.7 | 428.4 | 3.94x |
| matfull2_fwd | 4096 | 113.2 | 509.3 | 4.50x |
| matfull2_fwd | 16384 | 117.2 | 488.2 | 4.16x |
| matfull2_fwd | 65536 | 118.7 | 493.8 | 4.16x |
| matfull2_fwd | 262144 | 118.6 | 407.8 | 3.44x |
| matfull2_fwd | 1048576 | 117.9 | 334.9 | 2.84x |
| matfull2_fwd | 4194304 | 116.7 | 328.2 | 2.80x |
| matfull2_fwd | 16711744 | 115.5 | 323.2 | 2.80x |

</details>

## Correctness and testing

- The full CMake `ctest` suite passes; the only failures are ones that pre-date this branch and are unrelated to it.
- The SIMD paths are **not** bit-for-bit identical to the scalar paths (`-ffast-math` lets GCC call `libmvec`, whose x86_64 vector math functions are documented to within 4 ULP, so a result can differ from scalar `libm` by a few ULP).  This is normally irrelevant. The one test I ran that was sensitive to that (`grid_car3`) runs with `UseSIMD=0` so it stays bit-exact against its committed reference; the rest of the suite runs with SIMD enabled.
- Built and run clean under `-Wall`-style warnings and ASan/UBSan (no new warnings, no sanitizer findings) in addition to the normal `RelWithDebInfo` build.

## Caveats and limitations

- This is currently wired up and tested only for **GCC + glibc on x86-64**: the build gates on a GNU compiler and the `-march=x86-64-v3` flag, and `AST_ENABLE_SIMD` is off by default, so on any other toolchain it is simply a no-op.  Clang can almost certainly do the same (it also supports vector-ABI / `libmvec` math), but I haven't tested it with Clang at all yet.  glibc also ships a `libmvec` for aarch64 these days, but this build isn't set up or tested for that either.
- `astCPUCacheSize`'s CPUID path is x86-64; elsewhere it falls back to `sysconf` and then to conservative defaults (L1 32 KiB, L2 512 KiB, L3 8 MiB).
- Single-threaded, every transform is ultimately DRAM-bandwidth-bound at large enough N (the diagonal MatrixMap reaches that point almost immediately).  Going beyond that needs either threading (more memory bandwidth) or pipeline fusion so intermediate results are consumed while still cache-hot -- the latter is exactly what the end-to-end libasdf-gwcs path benefits from, and is left for future work.

### DIFF
--- a/.github/workflows/cmake.yaml
+++ b/.github/workflows/cmake.yaml
@@ -16,6 +16,7 @@ jobs:
         shared_libs: [ON, OFF]
         fortran: [OFF]
         sanitizers: [OFF]
+        simd: [OFF]
         include:
           # Add a Fortran build on each platform (shared only)
           - os: ubuntu-latest
@@ -23,22 +24,37 @@ jobs:
             shared_libs: ON
             fortran: ON
             sanitizers: OFF
+            simd: OFF
           - os: macos-latest
             build_type: Release
             shared_libs: ON
             fortran: ON
             sanitizers: OFF
+            simd: OFF
           # Debug builds with sanitizers
           - os: ubuntu-latest
             build_type: Debug
             shared_libs: ON
             fortran: OFF
             sanitizers: ON
+            simd: OFF
           - os: macos-latest
             build_type: Debug
             shared_libs: ON
             fortran: OFF
             sanitizers: ON
+            simd: OFF
+          # SIMD build (Ubuntu + GCC, Release). Exercises AST_ENABLE_SIMD.
+          # The code is compiled for x86-64-v3 (AVX2) with no runtime fallback,
+          # so it needs an AVX2-capable runner; GitHub's hosted x86-64 Linux
+          # runners (AMD EPYC) seem to provide AVX2 in practice; reference:
+          # https://runs-on.com/benchmarks/github-actions-cpu-performance/
+          - os: ubuntu-latest
+            build_type: Release
+            shared_libs: ON
+            fortran: OFF
+            sanitizers: OFF
+            simd: ON
 
     runs-on: ${{ matrix.os }}
 
@@ -62,6 +78,7 @@ jobs:
           -DBUILD_SHARED_LIBS=${{ matrix.shared_libs }}
           -DAST_BUILD_FORTRAN=${{ matrix.fortran }}
           -DAST_ENABLE_SANITIZERS=${{ matrix.sanitizers }}
+          -DAST_ENABLE_SIMD=${{ matrix.simd }}
           -DAST_C_STANDARD=11
 
       - name: Build
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -26,6 +26,7 @@ jobs:
           -DCMAKE_BUILD_TYPE=Debug
           -DBUILD_SHARED_LIBS=OFF
           -DAST_ENABLE_COVERAGE=ON
+          -DAST_ENABLE_SIMD=ON
           -DAST_C_STANDARD=11
 
       - name: Build
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,7 @@ include(CheckIncludeFile)
 include(CheckFunctionExists)
 include(CheckTypeSize)
 include(CheckCSourceCompiles)
+include(CheckCCompilerFlag)
 include(GNUInstallDirs)
 include(CMakePackageConfigHelpers)
 
@@ -33,6 +34,7 @@ option(AST_WITH_MEMDEBUG "Enable AST memory leak debugging" OFF)
 option(AST_ENABLE_WARNINGS "Enable extra compiler warnings for development builds" OFF)
 option(AST_ENABLE_SANITIZERS "Enable address/undefined sanitizers for supported compilers" OFF)
 option(AST_ENABLE_COVERAGE "Enable compiler coverage instrumentation for supported compilers" OFF)
+option(AST_ENABLE_SIMD "Enable SIMD/vectorization via GCC+glibc libmvec" OFF)
 option(AST_ENABLE_HUGE_TEST "Build the huge manual stress test converted from testhuge.f" OFF)
 option(BUILD_SHARED_LIBS "Build shared libraries" ON)
 option(BUILD_TESTING "Build the test program" ON)
@@ -125,6 +127,42 @@ if(AST_ENABLE_COVERAGE)
         )
     else()
         message(WARNING "AST_ENABLE_COVERAGE is ON, but no coverage profile is defined for compiler ${CMAKE_C_COMPILER_ID}")
+    endif()
+endif()
+
+set(AST_SIMD_FLAGS)  # per-file flags for SIMD-vectorized source files
+
+if(AST_ENABLE_SIMD)
+    if(CMAKE_C_COMPILER_ID STREQUAL "GNU")
+        check_c_compiler_flag(-fopenmp-simd    HAVE_FOPENMP_SIMD)
+        check_c_compiler_flag(-ffast-math      HAVE_FFAST_MATH)
+        check_c_compiler_flag(-march=x86-64-v3 HAVE_MARCH_X86_64_V3)
+
+        if(HAVE_FOPENMP_SIMD AND HAVE_FFAST_MATH AND HAVE_MARCH_X86_64_V3)
+            # -fopenmp-simd enables #pragma omp simd; -ffast-math is required
+            # for libmvec trig dispatch (sphmap.c) and improves vectorization
+            # of arithmetic loops (polymap.c).  -fno-associative-math is added
+            # back to avoid FMA-based reassociation changing other arithmetic.
+            list(APPEND AST_SIMD_FLAGS
+                -fopenmp-simd
+                -ffast-math
+                -fno-associative-math
+                -march=x86-64-v3
+            )
+            message(STATUS
+                "AST SIMD: GCC+libmvec vectorization enabled for sphmap.c, polymap.c, "
+                "matrixmap.c (${AST_SIMD_FLAGS})")
+        else()
+            message(WARNING
+                "AST_ENABLE_SIMD is ON but required flags are not available: "
+                "-fopenmp-simd=${HAVE_FOPENMP_SIMD} "
+                "-ffast-math=${HAVE_FFAST_MATH} "
+                "-march=x86-64-v3=${HAVE_MARCH_X86_64_V3}")
+        endif()
+    else()
+        message(WARNING
+            "AST_ENABLE_SIMD is ON but is only supported with GCC on glibc systems "
+            "(compiler: ${CMAKE_C_COMPILER_ID})")
     endif()
 endif()
 
@@ -750,6 +788,15 @@ set_target_properties(ast PROPERTIES
 )
 ast_apply_dev_options(ast)
 
+# Apply SIMD flags to specific source files only (not the whole library) so
+# that -ffast-math and -march= do not change the FP semantics of the rest of
+# libast. AST_HAVE_SIMD guards the restructured loops inside those files.
+if(AST_SIMD_FLAGS)
+    set_source_files_properties(src/sphmap.c src/polymap.c src/matrixmap.c
+        PROPERTIES COMPILE_OPTIONS "${AST_SIMD_FLAGS}")
+    target_compile_definitions(ast PRIVATE AST_HAVE_SIMD=1)
+endif()
+
 # libast references symbols (astGFlush, astPutErr_, etc.) that are provided
 # by the satellite libraries (ast_err, ast_grf_*). These are intentionally
 # unresolved in libast and resolved at final link time. Allow this.
@@ -901,6 +948,7 @@ if(BUILD_TESTING)
     ast_apply_dev_options(ast_test)
     add_test(NAME ast_test COMMAND ast_test)
     add_subdirectory(ast_tester)
+    add_subdirectory(perf)
 
     if(AST_ENABLE_COVERAGE)
         if(CMAKE_C_COMPILER_ID STREQUAL "GNU")
--- a/Makefile.am
+++ b/Makefile.am
@@ -30,7 +30,6 @@ GRP_C_ROUTINES = \
     src/lutmap.c \
     src/mapping.c \
     src/mathmap.c \
-    src/matrixmap.c \
     src/memory.c \
     src/moc.c \
     src/mocchan.c \
@@ -44,7 +43,6 @@ GRP_C_ROUTINES = \
     src/pointlist.c \
     src/pointset.c \
     src/polygon.c \
-    src/polymap.c \
     src/prism.c \
     src/ratemap.c \
     src/region.c \
@@ -56,7 +54,6 @@ GRP_C_ROUTINES = \
     src/specfluxframe.c \
     src/specframe.c \
     src/specmap.c \
-    src/sphmap.c \
     src/splinemap.c \
     src/stc.c \
     src/stccatalogentrylocation.c \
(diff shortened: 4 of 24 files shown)


dsberry:
This is all well outside my comfort zone, so I've no real comments to make other than it looks very promising.

timj:
Very impressive work. I had noodled around with MatrixMap SIMD vectorization but for 2x2 matrices I couldn't get it to make a difference and it wasn't clear we had bigger matrices. This is excellent.

embray:
> I had noodled around with MatrixMap SIMD vectorization but for 2x2 matrices I couldn't get it to make a difference.

The MatrixMap improvements are mostly marginal in practice I found.  SphMap was the biggest one by far--afterwards callgrind was still showing most of the remaining time spent on MatrixMap so I dove into it anyways, having become a little obsessed with chasing the last mile.  Now it looks like `TransformLoopSIMD` in `polymap.c` is back to being the biggest contributor but that's little surprise.

embray:
> This is all well outside my comfort zone, so I've no real comments to make other than it looks very promising.

Mine too honestly, I just tried to see what we could get almost "for free" from the compiler in terms of AVX2 instruction output, though it still required refactoring the kernels of each transform to vectorize efficiently over the input coordinate points.

I don't want to make a mess of the code though so if/when you have time I'd value your input on the overall architectural decisions (especially refactoring) and code style.  It's not urgent of course.

timj:
I am starting to think that for every AST test frameset we have in ast_tester (I know that directory needs to be classified and categorized into subdirectories but I think #37 will be made more complicated if I do that now) we should have a reference forward transform output (maybe of four corners?) so that we can compare outputs from changes like this to what we had before. We could store pixel_x, pixel_y, transform for 100 points in a CSV without blowing up the repo size. At least then you would know if you have substantially changed a mapping unexpectedly.

embray:
@timj That would not be a bad idea.  For what it's worth, as you can see, this didn't really have a big impact on the existing tests.  I don't think this would be a big problem for #37.  Even if more tests are added I can either incorporate those or, if we want to just focus on the cmake build that's fine too.

embray:
Hmm, as long as we're being persnicketty about satisfying the codecov report, there are several matrix cases that are uncovered by the tests, so I'll try to improve that too.

embray:
As far as I'm concerned, this is mostly ready--makes a nice difference already; also improved the tests, boosting coverage a bit.  There are two open questions remaining for me:

* [ ] Right now this enables SIMD automatically at runtime if explicitly enabled at build time via `cmake`, so users automatically get any benefits from it.  Questionable whether it should be enabled by default, or instead kept documented as experimental and "opt-in".  On the other hand, it still has to be explicitly opted into at build time so maybe that's good enough.
* [ ] Where and how to mention this in the documentation.

timj:
I imagine this hasn't been rebased since we added the transform oracle? That would help to solidify whether the SIMD is working properly. I imagine we would also need a way to ensure that we generate oracles from the non-SIMD path? Should we do it as experimental first with an environment variable gate?

embray:
I think I have tested it against that, since I did make a private branch with this merged in.  But I'll double-check.  Needs rebasing anyways.

embray:
@timj  Rebased, and confirmed that the oracle tests are passing.  Also confirmed via a coverage build that my new code paths were being exercised in the test.

I also re-ran my benchmarks and they were consistent with previous results, with some apparent slight improvement in the polynomial evaluations, and a slight regression in the non-diagonal matrix benchmarks.  I haven't looked into whether there are any code changes that could account for that. But I was running the benchmark just on my laptop with my system under average load and performance settings so it's not a clean benchmark; just a sanity check.

PR #53, opened 2026-06-04
Handle DBL_DECIMAL_DIG inconsistency

embray:
This came up at #38: https://github.com/Starlink/ast/pull/38#issuecomment-4273909152

But was never handled.  I agree that just setting the fallback `DBL_DECIMAL_DIG` to be consistent between C99 and C11 is least disruptive, and this fixes all the test failure I was having on my machine where the compiler was defaulting to C99.

Also fixed several build warnings that were coming from the Claude-generated tests (warnings that were enabled by default in GCC but maybe not in clang).


timj:
Thank you for doing this.

PR #55, opened 2026-06-16
Add astGetKeyMap protected method to Object

### LINKED ISSUE #54: Allow AST Object attached as proxy to be copied
When attaching a "proxy" object to an `AstObject` using `astSetProxy`, if the attached object is later cloned with `astCopy` the proxy object is not cloned and is set to `NULL` on the copy.

This makes sense of course--it takes an arbitrary pointer which AST treats opaquely. It wouldn't know what to do with it on `astCopy`.

My proposal is that *if* `astSetProxy` is called with a pointer to, or really a handle to (setting a raw pointer here would have its own dangers) an `AstObject` of any kind, then the deepcopy returned by `astCopy` includes the proxied `AstObject` (e.g. an `AstKeyMap` or otherwise).

This is a little tricky to get right between the public and protected APIs though.

Another alternative might be extending the API to allow the `AstObject` to own the pointer to the proxy object, and register copy and destroy functions for it. This approach might be more flexible on the whole, but not clear if it's needed.

Motivation is to simplify some of the code in `yamlchan.c` to allow mapping objects to carry information about the transform from which they were deserialized--currently this is not possible because setting a proxy on them doesn't survive `astCopy` (simplification is another matter too but that can be avoided more easily).

Yet another alternative would be to add a new field on `AstMapping` or even `AstObject` to hold an optional keymap that can be used as an internal scratch area (distinct from the publicly accessible attributes).

I'm working on an implementation of the first approach (allowing a proxied `AstObject` to be deepcopied) but wanted to register the issue in case you all have any thoughts about it.

embray:
Resolves #54.  My motivation here is primarily for use in yamlchan.c to make round-tripping certain transforms easier.

Currently it uses a lot of structural pattern-matching (e.g. `FindSphericalCartesian`) to reconstruct the appropriate ASDF transforms from a slice of the mapping list.  This code *is* still useful and necessary -- e.g. for hand-constructed astMapping, or one read from another source, one will still need to figure out the most appropriate way to serialize it back to ASDF.

But for the case of deserializing from ASDF it will also be very useful to maintain additional information on the deserialized mappings for accurate round-tripping without having to go down the full `Find<Foo>` path.  This idea is already used, to an extent, by the `KeyMap` attached to some mappings via `astSetProxy`, but that proves not as useful since the proxy object can't survive copying, per the discussion in #54.  So this would be used to replace that.

Had started out with trying to reorganize `ast_tester/testobject.c` in anticipation of adding more unit tests for this, but then realized since it's a protected interface it needs to be compiled with `astCLASS` defined, so ended up having to put the tests for this in a separate test program anyways.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -921,7 +921,9 @@ if(BUILD_TESTING)
                         --ignore-errors negative
                         --rc branch_coverage=1
                     COMMAND ${LCOV_EXECUTABLE}
-                        --extract "${AST_COVERAGE_FULL_INFO}" "${CMAKE_SOURCE_DIR}/src/*"
+                        --extract "${AST_COVERAGE_FULL_INFO}"
+                            "${CMAKE_SOURCE_DIR}/src/*"
+                            "${CMAKE_SOURCE_DIR}/ast_tester/*"
                         --output-file "${AST_COVERAGE_INFO}"
                         --ignore-errors unused,negative
                         --rc branch_coverage=1
--- a/ast_tester/CMakeLists.txt
+++ b/ast_tester/CMakeLists.txt
@@ -156,6 +156,7 @@ ast_add_test(testresimp INTERNAL_HEADERS)
 ast_add_test(testaxis)
 ast_add_test(testframe)
 ast_add_test(testunitnorm)
+ast_add_test(testobjectkeymap)
 
 # --- Tests needing data files ---
 configure_file(2dspline_c.dat "${CMAKE_CURRENT_BINARY_DIR}/2dspline_c.dat" COPYONLY)
--- a/ast_tester/testobject.c
+++ b/ast_tester/testobject.c
@@ -1,3 +1,5 @@
+/* Test program for the AstObject base class. */
+
 #include "ast.h"
 #include <libgen.h>
 #include <stdio.h>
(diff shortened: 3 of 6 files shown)


dsberry:
A few points about your code changes:

- You've made it a static function (like, say, astTune) rather than a virtual function (like, say, astCopy). This means it can't be extended or replaced by any other class of AST object. That's probably okay for the moment, but it's not inconceivable that a subclass may want to replace or extend it at some time in the future. Just something to be aware of.
- The current version of astGetKeymap returns the pointer from the class structure directly, meaning that it's imperative that calling code does not annul the returned pointer. This is a bit dangerous because  AST in all other (I think) circumstances returns a cloned pointer and expects the caller to annul it. Having one place that is different is likely to cause confusion. For instance, look at the GetFrame method in frameset.c, which uses astClone to return a cloned pointer. In general, AST doesn't use  "borrowed" pointers derived from direct pointer assignment. For consistency, I'd return "astClone(this->keymap)" and change the prologue documentation to say the pointer should be annulled using astAnnul when no longer needed.
-  Just a reminder that in AST "cloning" a pointer refers to returning a new reference to an existing object.
-  Prior to the addition of "keymap" the Object class structure contained no references to other Objects, unlike say FrameSet that contains many references to other Objects. So we now need to include the keymap in the GetObjSize and ManageLock functions in object.c. You could use the versions of these functions in frameset.c to see how the code should look.

Otherwise, it all looks good.

embray:
Huh, to my mind returning a borrowed reference seemed to make more sense / be more convenient, so I documented it as such.  But if it would be inconsistent with how literally everything else works I'm happy to change that.

Not sure why a subclass would want to extend this--maybe to provide a default, non-empty keymap?  Would be easy enough to change later if needed.

I'll look at GetObjSize and ManageLock.

embray:
Should be better now

- `astGetKeyMap` returns a new reference via `astClone`
- `GetObjSize` and `ManageLock` accounted for

The more I dig into the internals of how this library works the more impressed I am with how well it's designed; it's really a pleasure to work with.  Making each Object hierarchy uniquely owned by a single thread is a good call.

embray:
Oh, forgot to mention, I also added `ast_tester/*` to the paths extracted by lcov for the coverage report.  This way when running the coverage report locally I can also get coverage on the tests which is very useful.

I was for a while scratching my head about why codecov's report includes all sources, even those that aren't included in the uploaded lcov.info.  Turns out codecov's tool actually runs over the entire build directory automatically, extracting any coverage information it can find, including raw gcov outputs, and incorporate that.  That can be disabled by passing it `disable_search: true` in the codecov-action configuration, but I left that out for now. 

But by `ast_tester` code through lcov we can also use things like `LCOV_EXCL_LINE` on lines that should *not* be reached if the tests are passing, improving overall line coverage.

embray:
Updated to include an `astHasKeyMap` companion method, which can be used to check if an Object has an associated KeyMap without creating an empty one on the fly as `astGetKeyMap` does (working on integrating this into yamlchan).

I'm still having some doubts as to whether it's even the right call to include this in the `astDump` serialization, since it's only intended right now for internal use, and maybe shouldn't leak into the serialization format.

dsberry:
Your call. The default assumption would be that it is dumped, along with every other aspect of the Object. So if you decide not to include it in the dump, then it should be very clearly documented somewhere.

embray:
Right.  I think that's why it's best to keep for now.

embray:
Is this OK to merge?  I've also tried it out a bit in #66 and it will help with some other transforms I'm adding support for.

embray:
(P.S. @timj I saw you gave me merge permissions anyways, but I want to make sure not to step on any toes)

dsberry:
Merging is okay with me

PR #57, opened 2026-06-18
polygon: Fix bug in astMask when using very large polygons on the sky

### LINKED ISSUE #56: `astMask` and `astTranN` disagree for a SkyFrame `Polygon` whose interior is the larger lobe
I was trying to get some masking code to work and got confused. I asked Claude to analyze the situation and it thinks there is a bug, or at least an undocumented inconsistency.

[testpolygonmask.c](https://github.com/user-attachments/files/29055347/testpolygonmask.c)

## Summary

A `Polygon` defined in a `SkyFrame` divides the sphere into two lobes.
When the vertices are wound so that the *un-negated* interior is the **larger** lobe, `astMask` and the region's own membership test (`astTranN` / `astPointInRegion`) disagree: points that `astTranN` reports as inside the region are **not** masked by `astMask`.
In the reproducer below, `astMask` masks 0 of 441 grid pixels even though every one of those pixel centres is reported inside the region by `astTranN`.

Obtaining the *same logical region* by negating a small-lobe polygon (`astSetI(poly,"Negated",1)`) masks correctly, so the problem is specific to the winding-selected large lobe, not to large-lobe regions in general.

## Versions

- Observed with AST 9.3.0 (via starlink-pyast) and AST 9.3.1 (built from the current `master`/source tree).
- Platform: macOS (arm64); not believed to be platform-specific.

## Minimal reproduction (C)

See the companion file `testpolygonmask.c` (written in the style of `testregions.c`).
Built against `libast` it currently prints:

```
Error: astMaskI masked 0 of 441 in-region pixels (astTranN reports all of them inside)
Polygon-mask tests failed
```

The test:

1. builds a small square near the equator in an ICRS `SkyFrame`, wound so the un-negated interior is the large lobe;
2. uses `astTranN` to confirm the square's centre is **outside** and a point ~0.1 rad away is **inside** (i.e. the interior really is the large lobe);
3. masks a 21×21 pixel grid centred on that far point — a patch lying entirely inside the large lobe — and asserts that all 441 pixels are masked.

Step 3 currently masks 0 pixels.

## Minimal reproduction (Python / starlink-pyast)

```python
import numpy as np
import starlink.Ast as Ast

sky = Ast.SkyFrame("System=ICRS")
ra0, dec0, half = 0.5, 0.0, 0.001                      # radians, near the equator

# Vertices wound so the un-negated interior is the LARGE lobe.
verts = np.array([[ra0-half, ra0+half, ra0+half, ra0-half],
                  [dec0-half, dec0-half, dec0+half, dec0+half]])
poly = Ast.Polygon(sky, verts)
assert poly.pointinregion([ra0, dec0]) is False        # square centre: outside
assert poly.pointinregion([ra0 + 0.1, dec0]) is True   # far point:     inside

# A 21x21 grid ~0.1 rad from the square, entirely inside the large lobe.
fra, fdec, g, span = ra0 + 0.1, dec0, 20, 0.002
sky2pix = Ast.WinMap([fra-span, fdec-span], [fra+span, fdec+span], [0, 0], [g, g])

xs = np.linspace(fra-span, fra+span, g+1)
ys = np.linspace(fdec-span, fdec+span, g+1)
assert all(poly.pointinregion([x, y]) for x in xs for y in ys)   # all inside

arr = np.zeros((g+1, g+1), dtype=np.int64)
poly.mask(sky2pix, 1, [0, 0], [g, g], arr, 1)
print("masked", int(arr.sum()), "of", (g+1)**2)         # -> masked 0 of 441
```

For contrast, the same region built as a negated small-lobe polygon masks all 441:

```python
poly2 = Ast.Polygon(sky, verts[:, ::-1])   # other winding -> small lobe interior
poly2.negate()                             # now interior is the large lobe
# ... same grid ... -> masks 441 of 441
```

## Analysis

The behaviour appears to stem from three interacting facts in the current source.

1. **The constructor cannot normalise the winding on a SkyFrame.**
   `polygon.c` (changelog `6-DEC-2013`) reverses the vertices at construction "if necessary, to ensure that the unnegated Polygon is bounded," because "the parent Region class assumes that unnegated regions are bounded."
   But `Polygon`'s `GetBounded` returns non-zero unconditionally for any SkyFrame:

   ```c
   /* Regions defined within SkyFrames (i.e. on a sphere) are always bounded,
      since a finite region has a finite negation. */
      if( astIsASkyFrame( bfrm ) ) {
         result = 1;
   ```

   (this is the intended behaviour, cf. changelog `9-JUN-2021`: "Regions on SkyFrames are all bounded").
   Consequently the "is the un-negated polygon bounded?" test is always true on a SkyFrame, the auto-reversal never fires, and a polygon whose un-negated interior is the larger lobe is left as supplied.

2. **`astGetRegionBounds` ignores `Negated`**, returning the bounds of the vertex outline (its own documentation: "The value of the Negated attribute is ignored (i.e. it is assumed that the Region has not been negated)").

3. **`astMask` assumes the interior is enclosed by that bounding box / boundary mesh.**
   In the `MAKE_MASK` macro in `region.c`, after obtaining the region bounds it states:

   ```c
   /* All points outside this box are either all inside, or all outside, the
      Region. So we can speed up processing by setting all the points which are
      outside the box to the supplied data value (if required). ... */
   ```

   and only evaluates the per-point `Transform` for pixels inside that box, filling the remainder by assumption.
   This is valid only when the region's interior is the lobe enclosed by the vertex outline.
   For the winding-selected large lobe the interior extends far beyond the vertex outline, so pixels in the interior but away from the outline are treated as exterior — disagreeing with `astTranN`, which evaluates the geometry per point and is correct.

The net effect: `astTranN`/`astPointInRegion` are reliable for these polygons, but `astMask` is not.

## Expected vs actual

- **Expected:** for any Region, `astMask(...,inside=1,...)` sets exactly those pixels whose centres `astTranN`/`astPointInRegion` report as inside.
- **Actual:** for a SkyFrame `Polygon` whose (un-negated) interior is the larger lobe, `astMask` masks only pixels near the vertex outline (here, none), missing the bulk of the interior.

## Possible directions

Not prescriptive — flagging where a fix might sit:

- make `Polygon`'s `GetBounded`/winding handling on a SkyFrame distinguish the two lobes so the constructor can normalise (or record) which lobe is the interior; and/or
- have `astMask`/`astResample` not assume "outside the bounding box ⇒ outside the region" when the region's interior is the unbounded-by-outline lobe (i.e. consult the same predicate `astTranN` uses).

A consistency check that `astMask` and `astTranN` agree on a sample of points would have caught this.

dsberry:
Fix bug that caused some Polygons defined on the sky to be misinterpreted by astMask. Previously, a Polygon that was supplied such that the unnegated Region represent more than half the sky was effectively inverted by astMask (but not by astTranN). The fix was to reverse the order of the supplied vertices and then set the Invert attribute when constructing the Polygon, if the supplied Polygon represents more than half the sky.

Fixes #56

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -88,7 +88,30 @@ and `__LINE__` captures instead of hardcoded line numbers.
 
 Fixed `testsplinemap_c.c` to check `fscanf` return values.
 
-Fixed `testthreads.c` to return NULL from `worker()` thread function.
+Fixed `testthreads.c` to return NULL from `worker()` thread function. Later
+reworked so each worker watches its own thread-local status and reports the
+outcome back through the shared struct, replacing the Starlink EMS calls
+(`errStat`/`errAnnul`) that are no-ops in the standalone build and so could
+never detect the expected lock error.
+
+### Exit-status audit
+
+ctest decides pass/fail purely from each program's exit code (no pass/fail
+regex), so a test that always returns 0 reports success even when its own
+checks fail. Eight programs ran off the end of `main()` with no return
+statement (an implicit `return 0` in C99) and so had been silently passing:
+`testerror`, `testobject`, `testaxis`, `testframe`, `testunitnorm`,
+`testsplinemap_c`, `testthreads`, `testyamlchan` (and `testregions`, which
+returned a hardcoded 0). All now `return astOK ? 0 : 1`. Making the status
+real exposed two genuine failures that had been masked:
+
+- **testyamlchan**: its input data files (`imaging_wcs.asdf`, `tanSipWcs.txt`,
+  `lsst_wcs.txt`) existed in the source tree but were never copied into the
+  build directory; added the missing `configure_file` calls. Running the test
+  under AddressSanitizer then exposed two pre-existing buffer overflows in
+  `yamlchan.c`/`memory.c` (stack overrun in `ReadPoly`'s `dimd`/`dimw`, and a
+  `str - 1` underread in `astChrSplit_`), fixed by cherry-picking the upstream
+  fixes from PR #39.
 
 ## Phase 2 conversion approach
 
@@ -131,7 +154,7 @@ Key issues:
 
 - **testfitschan.c**: FITS card padding ignored during assertions to match Fortran string comparison rules. Fixed a `heap-use-after-free` bug in `astStore_` (memory.c) exposed by AddressSanitizer during `astConvert`.
 
-- **testregions.c**: Translated automatically and then fixed up to cast `astTranN` input arrays to `(const double *)`.
+- **testregions.c**: Translated automatically and then fixed up to cast `astTranN` input arrays to `(const double *)`. Several automatic-translation defects were later corrected against `testregions.f`: the per-section "X tests failed" messages were printed unconditionally (the Fortran `if status != OK` guard had been dropped); `checkBox`/`checkCircle`/`checkEllipse` each used the wrong FITS header (a single unrelated block had been copied into all three); `checkdump` compared dump strings with `strcmp` instead of falling back to `astOverlap()==5` as the Fortran does; and four `frm1 = astFrame(...)` statements had been absorbed into `//` comments. The `astGetRegionDisc` check in `checkCmpRegion` uses tolerances relaxed to `1e-7` (from the Fortran's `1e-9`/`1e-8`) because that disc is fitted to a subsampled boundary mesh and so varies at the `~1e-8` level with build/optimisation settings; the Fortran test built against the autoconf AST passes the tight tolerance on the same machine.
 - **testrebinseq.c**: Translated to use `astRebinSeq[I|F|D]` depending on the types of the in/out pointers.
 
 - **testrebin.c**: Collapses 27 Fortran subroutines (9 test groups × 3 type
--- a/ast_tester/CMakeLists.txt
+++ b/ast_tester/CMakeLists.txt
@@ -382,6 +382,11 @@ endif()
 
 # --- Conditional tests ---
 if(HAVE_YAML)
+    # Input data files read by the round-trip tests (the .asdf/.yaml files
+    # written by the test are created in the working directory at run time).
+    configure_file(imaging_wcs.asdf "${CMAKE_CURRENT_BINARY_DIR}/imaging_wcs.asdf" COPYONLY)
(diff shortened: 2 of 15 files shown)


dsberry:
@timj Not sure what else needs doing on this one. The test program created by claude now passes, as do all the other tests. Can I leave you to merge the PR and close the issue?

timj:
Thank you. I will take a look and merge.

Claude has found a few other bugs as well that I will try to do PRs for.

timj:
I added the failing test and then realized that it wasn't failing. Turned out some C tests weren't returning bad status. Fixing that led to some tests failing because of bad Fortran to C conversion. Fixing those led to me needing the overrun fixes from #39 (@embray this will cause a rebase issue for you with the prolog -- sorry).

embray:
@timj Not a problem.  See also my comment here: https://github.com/Starlink/ast/pull/55#issuecomment-4729740683  I could maybe also pull that change out to a separate PR but since that one is ready as far as I'm concerned maybe it can stay there.

A few other tests have suspiciously low line coverage.  A lot of that is just due to failure conditions that should be excluded from the coverage analysis.  Though a few, especially testregions.c, don't seem to be running properly at all.  I'm going to take a look at that.

timj:
Didn't this PR turn on the regions tests that were all failing previously?

dsberry:
You've lost me there I'm afraid. Which region tests? And I'm not sure what you mean by turning them on.

timj:
I am talking about testregions.c (The port of the fortran) which was not checking status properly and so a test was failing early on (because of a porting error) and then no other tests were running with the test returning good status overall.

timj:
I also only just noticed that testpolygonmask.c was added on this PR but was not added to the CMake test build.

dsberry:
Sorry Tim, I'm not sure what's going on here. As far as I can remember all I did on this PR was to commit the original fix in polygon.c.

timj:
Sorry. My fault

* I committed testpolygonmask.c by mistake and didn't wire it up to any tests. I had forgotten it was there.
* I committed the test code to testregions.c
* I realized that testregions.c was broken because it was passing the test on master when it should be failing.
* It turned out that status wasn't being returned to shell.
* Then I realized that status wasn't being handled at all and an earlier region test was failing.
* This led to me finding other tests that weren't setting status properly.
* Everything then worked but the testpolygonmask.c was orphaned. I will try to remember to remove it on a new PR.

PR #58, opened 2026-06-20
Twelve bugs found by Claude

timj:
@dsberry Whilst noodling around with Claude it seems to have found numerous bugs which seem to be correct to my eyes. It would be great if you could take a look at some point.

The VELOSYS precision issue was fixed in #59.

# libast bug fixes

A set of correctness fixes to `libast`, found by a systematic audit and each verified against the current source. Eleven genuine defects are fixed (one commit each) across `src/`, with regression tests and any forced reference fixture updates included alongside the relevant fix. The final commit documents a twelfth, related precision issue that is deliberately left unchanged pending maintainer input (see the end of this document).

All work was validated with the CMake developer build (`-DAST_C_STANDARD=11 -DAST_ENABLE_WARNINGS=ON -DAST_ENABLE_SANITIZERS=ON`); the full test suite (743 tests) passes and the changes add no new warnings or sanitizer failures.

### 1. `DateObs` drops leading zeros in DATE-OBS fractional seconds
**File:** `src/fitschan.c` (`DateObs`)

The fractional-seconds field of an ISO-8601 `DATE-OBS` string was read with `%d` into an `int`, discarding leading zeros and so losing the field width. The digit-reconstruction loop then produced the wrong magnitude: `"2017-07-21T08:38:39.087"` parsed as `87` and reconstructed to `0.87 s` instead of `0.087 s` — a 0.783 s error stored on the FrameSet's `Epoch`. The field width is now captured with `%n` markers bracketing the `%d` (both the plain and trailing-`Z` patterns) and the fraction recovered as `fsc / pow(10, width)`. Regression test added; `azel` reference fixtures corrected.

### 2. `SIPIntWorld` reads `nout` from `astGetNin`
**File:** `src/fitschan.c` (`SIPIntWorld`)

`nout` was initialised from `astGetNin(map)` instead of `astGetNout(map)`. Harmless when `Nin == Nout`, but for a non-square sub-mapping the output count was wrong and the downstream per-output logic indexed past the real output count. One-line fix.

### 3. `AxAngle` nudge reads a stale loop index
**File:** `src/frame.c` (`AxAngle`)

When the reference point's component on the measured axis is zero but the offset point's is non-zero, the degeneracy-breaking nudge was scaled by `fabs(b[iaxis-1])`, where `iaxis` is the component-loop counter left at `naxes` after the loop — so it read `b[naxes-1]` rather than the intended `b[axis-1]`. For configurations such as `a=(0,0), b=(1,0)` on a 2-D Frame the nudge became zero, collapsing the offset point onto the reference point, and `astAngle` returned `AST__BAD` instead of a valid position angle. Regression test added.

### 4. SpecMap velocity helpers cast RA/Dec to `float`
**File:** `src/specmap.c` (`LsrkVel`, `LsrdVel`, `LgVel`, `GalVel`)

The LSRK/LSRD/LG/Galactic velocity helpers cast the source RA/Dec down to `float` before calling `palRvlsrk`/`palRvlsrd`/`palRvlg`/`palRvgalc`. Those PAL routines are double precision throughout; the casts were a hangover from the single-precision SLALIB `REAL` signatures and discarded ~20 mas of angular precision, perturbing the computed velocities by ~1 mm/s. The sibling helpers (`BaryVel`, `GeoVel`, `TopoVel`, `UserVel`) already pass `double` straight through. Removing the casts shifts FITS-CLASS `VELO-LSR`/`VLSR` by ~1.3 mm/s toward the full-precision value; the affected reference fixtures are regenerated.

### 5. `astDat` inverse divisor wrong for the pre-1972 rubber-second era
**File:** `src/timemap.c` (`astDat_`, inverse branch)

The pre-1972 TAI-UTC model is piecewise linear, `dTAI = a + b·(utc − t0)`, with the slope `b` in seconds/day and `astDat` returning seconds. Inverting the day-space relation `tai = utc + dTAI/SPD` and solving for `UTC − TAI` at a given TAI gives a divisor of `(1 + b/SPD) ≈ 1.00000003`. The inverse branch instead used literals that dropped the `/SPD` factor entirely (`1.001296`, `1.0011232`, i.e. `1 + b`), and the 1968-era lines used `1.02592` (`1 + 10·b`); one 1965 line even carried an anomalous `1.01296`. All were wrong by a units factor and leaked ~164 ms (~1.9e-6 d) on a `UTC → TAI → UTC` round trip near MJD 40000. Each divisor literal is replaced by the computed form `(1.0 + slope/SPD)` so the slope and divisor cannot drift apart. Regression test asserts the round trip at MJD 40000 closes to within 1e-9 d.

### 6. `astRate` returns `AST__BAD` at `x0 = 0` for narrow-domain Mappings
**File:** `src/mapping.c` (`Rate`)

`Rate` seeds the gradient-search interval at a small fraction of `x0`, but fell back to `1.0` when `x0` was exactly zero. The working-interval loop immediately multiplies the seed by ten, so the first sampled interval spanned roughly `[-5, 5]`; for a Mapping defined only over a narrow domain (e.g. a `ChebyMap` on `[-1, 1]`) every sample lay outside the domain, `FindGradient` always failed, and `astRate` returned `AST__BAD` even though the derivative at `x0 = 0` is well defined. The `x0 = 0` case now seeds the same small step used for the non-zero case; the existing grow-by-ten loop still escalates the interval when needed, so unrestricted Mappings are unaffected. Regression test added (`f(x) = 3x` over `[-1, 1]` returns `3.0` at both `x0 = 0.5` and `x0 = 0.0`).

### 7. Out-of-bounds read in `ScalePolyInputs`
**File:** `src/fitschan.c` (`ScalePolyInputs`)

The inverse-coefficient loop scaled each coefficient by `scale[iin]`, reusing the `scale` pointer the forward-coefficient loop had advanced to one past the end of the `scales` array (`scales + nin`). For a PolyMap with an explicit (non-iterative) inverse this read `scales[nin + iin]`, out of bounds. The intended index is `scales[iin]` (per the loop comment). Latent on the SIP-write path, which always supplies a PolyMap with `IterInverse` set (so the loop body never runs), but undefined behaviour for any caller passing an explicit-inverse PolyMap.

### 8. `CLASSFromStore` multiplies spectral CDELT by `specfactor` twice
**File:** `src/fitschan.c` (`CLASSFromStore`)

`cdelt[axspec]` is multiplied by `specfactor` in place when converting the stored spectral CDELT to Hz, but it was then multiplied by `specfactor` a second time when forming the neighbouring-channel input to the frequency→velocity Mapping. `crval[axspec]` is not pre-scaled, so its scaling is correct; only the CDELT term was doubled (which would scale DELTAV by the square of the unit factor). Dormant because `AddEncodingFrame` forces the SpecFrame Unit to `"Hz"` before the encoder runs (`specfactor == 1`), but wrong for any non-Hz spectral unit. No reference fixtures change.

### 9. Uninitialised read in `MatrixMap` `MapSplit`
**File:** `src/matrixmap.c` (`MapSplit`)

When classifying which output rows depend only on the selected inputs, `MapSplit` treats a BAD matrix element the same as zero (no dependency). A kept row could therefore reach the element-copy loop with a BAD value in a selected column; that loop advanced the output pointer only for good elements, leaving a gap, and the `MatrixMap` constructor then read exactly `nin*nout` elements — consuming the uninitialised slot and producing a non-deterministic sub-matrix. Such a row cannot be cleanly split out (the output genuinely depends on a BAD input), so the split is now rejected in that case, matching the existing rejection of rows that mix selected and unselected inputs. Real WCS matrices (`CDi_j`/`PCi_j`) are always fully populated with finite values, so no shipped fixture exercises this path.

### 10. Stale-slot read in `astRate` zero-range confirmation
**File:** `src/mapping.c` (`Rate`)

The "is the previous interval also zero-range?" check in both the expanding and contracting interval loops indexed the range array by the loop counter (`iin`), but the gradient range/mean are stored at the running indices `itop`/`ibot`, which advance only when `FindGradient` succeeds. When an interval is skipped, `iin` and `itop`/`ibot` diverge, so `y[iin-1]`/`y[iin+1]` reads a slot never written for the current search. Indexed by the store index (`itop-1`/`ibot+1`) instead. Latent for the smooth Mappings in the test suite, so no fixtures change.

### 11. `PermMap::Equal` skips constant-value comparison for matching indices
**File:** `src/permmap.c` (`Equal`)

`Equal` compared the referenced constant *values* only inside an "if the permutation entries differ" branch. When two PermMaps fed an input or output from the *same* constant slot (`p1 == p2 < 0`), that branch was skipped, so they were declared equal without inspecting the stored constants — two PermMaps with `inperm = [-1]` but constants `{5.0}` vs `{7.0}` compared equal, which could cause an over-merge in `astSimplify`. The comparison is restructured so that whenever both entries are negative the constant values are compared regardless of whether the indices match; the non-negative case still requires an exact index match. A NULL constants array is treated as supplying `AST__BAD`, matching the forward transformation. Regression test locks all four cases (same/different index × same/different value, plus surplus unreferenced constants).

### 12. SpecFluxFrame: identical frames now compare equal (`src/specfluxframe.c`)

`Frame::Equal` is conversion-based — two Frames are equal iff `astConvert` between them simplifies to a `UnitMap`. The SpecFlux↔SpecFlux conversion built by `MakeSFMapping` always contains a flux-scaling `RateMap` (the rate of change of the output spectral system with respect to the input, used to rescale the flux density), which `astSimplify` cannot reduce to a `UnitMap` even when the target and result are identical. As a result the conversion stayed a `CmpMap` and `astEqual` returned 0, so a `SpecFluxFrame` did not compare equal to a separately-constructed identical `SpecFluxFrame`, nor to a copy of itself (except via the same-pointer short-circuit in `astConvert`); the component `SpecFrame` and `FluxFrame` each compared equal correctly, so only the compound frame was affected. `MakeSFMapping` now returns a 2-input `UnitMap` directly when the spectral components are equal and the flux components are equal (both component comparisons are themselves conversion-based and reduce correctly), bypassing the `RateMap` construction; a regression test in `testspecflux.c` covers identical frames, a copy, and frames with differing flux systems. Note that this changes `astEqual`/`astSimplify` behaviour for SpecFluxFrames — the full suite passes, but reviewers may wish to confirm the short-circuit is placed where they'd prefer.


src/fitschan.c
```diff
@@ -21479,7 +21512,17 @@ static AstFrameSet *MakeFitsFrameSet( AstFitsChan *this, AstFrameSet *fset,
...
+
+   Note: this astFormat/astSetC round trip quantises the full-precision
+   fiducial RA/Dec to the SkyFrame's default formatting precision (~0.1s
+   of time / 1 arcsec) before storage, which degrades the VELOSYS value
+   later derived from RefRA/RefDec by ~tens of mm/s. Setting RefRA/RefDec
+   to the full-precision radian values directly would remove this, but is
+   not as simple as it looks (the FrameSet astSetD route mis-parses the
+   radian value via the sexagesimal "refra=" parser, and the typed-setter
+   route skips the integrity remap astSetC triggers). Left as-is pending a
+   decision from the AST developers. */
```
timj: Would temporarily setting Digits to a high value for the astFormat fix this?

dsberry:
We definitely need to retain the FrameSet integrity checking, so I think the best fix is just to increase the precision of the formatting temporarily before using the existing code to copy the RefRA  and RefDec values. See PR  #59 .

ast_tester/azel.native
```diff
@@ -28,7 +28,7 @@ BEGAST_E= 'SkyFrame'           / Description of celestial coordinate system
 IDENT_A = '" "     '           / Permanent Object identification string         
 ISA_A   = 'Object  '           / AST Object                                     
 NAXES_B =                    2 / Number of coordinate axes                      
-EPOCH_A =   2017.5519807923772 / Julian epoch of observation                    
+EPOCH_A =   2017.5519807675655 / Julian epoch of observation
```
embray, line 31: 😬

ast_tester/testtime.c
```diff
@@ -538,6 +538,43 @@ int main( void ) {
...
+    * inverse divisor used to leak ~164 ms (~1.9e-6 d); a wrong intercept
+    * constant in the 1966 inverse branch leaked ~0.1 s at MJD 39500. */
+   {
+      /* UTC MJD inside each era, and the expected TAI-UTC there (seconds). */
+      double era_utc[] = { 40000.0, 39500.0, 39060.0, 38970.0, 38880.0,
+                           38790.0, 38700.0, 38560.0, 38440.0, 38360.0,
+                           38000.0, 37590.0, 37400.0, 37200.0 };
+      double era_dt[]  = { 6.478578, 5.282578, 4.227634, 4.010994, 3.794354,
+                           3.577714, 3.361074, 3.079634, 2.824114, 2.726482,
+                           2.222130, 1.748658, 1.552418, 1.288218 };
```
embray, line 557: Out of due diligence I cross-checked these results against `astropy.time` and it at least matches up there:

```python
In [45]: times_mjd_utc = Time([40000.0, 39500.0, 39060.0, 38970.0, 38880.0,
    ...:                      38790.0, 38700.0, 38560.0, 38440.0, 38360.0,
    ...:                      38000.0, 37590.0, 37400.0, 37200.0], scale='utc', format='mjd')

In [46]: (times_mjd_utc.tai.value - times_mjd_utc.value)*86400.0
Out[46]:
array([6.47857769, 5.28257797, 4.22763373, 4.01099401, 3.79435429,
       3.57771395, 3.36107423, 3.07963402, 2.82411401, 2.72648202,
       2.22213012, 1.74865779, 1.55241815, 1.28821784])
```

src/fitschan.c
```diff
@@ -1350,6 +1350,33 @@ f     - AST_WRITEFITS: Write all cards out to the sink function
 *        values being formatted with the default number of digits (7) and
 *        then unformatted. Now, the Digits value is temporarily increased
 *        to 20 before formatting these values.
+*     20-JUN-2026 (TIMJ):
+*        Fix DateObs: the fractional seconds field was read with "%d"
```
timj, line 1354: These changes all look fine to me.

src/frame.c
```diff
@@ -1441,7 +1448,7 @@ f     AST_DISTANCE function.
             aa[ axis - 1 ] += 10000.0*DBL_EPSILON*fabs( a[ axis - 1 ] );
 
          } else if( b[ axis - 1 ] != 0.0 ) {
-            aa[ axis - 1 ] = 10000.0*DBL_EPSILON*fabs( b[ iaxis - 1 ] );
+            aa[ axis - 1 ] = 10000.0*DBL_EPSILON*fabs( b[ axis - 1 ] );
```
timj, line 1451: Looks correct at first glance.

src/mapping.c
```diff
@@ -9068,8 +9092,11 @@ static double Rate( AstMapping *this, double *at, int ax1, int ax2,
    interval also had zero range. Otherwise, it's probably just a numerical
    fluke. If the previous interval also had a range of zero, we can forget
    the rest of the algorithm since the supplied transformation is linear
-   and we now have its gradient. So leave the loop. */
-               } else if( range == 0.0 && y[ iin - 1 ] == 0 ) {
+   and we now have its gradient. So leave the loop. Index the range array
+   by the store index "itop" (not the loop counter "iin"), since "itop"
+   only advances when a value is actually stored - the two diverge if any
+   interval was skipped because FindGradient failed. */
+               } else if( range == 0.0 && y[ itop - 1 ] == 0 ) {
```
timj, line 9099: The explanation for using itop and ibot seems reasonable.

src/permmap.c
```diff
@@ -96,6 +96,15 @@ f     The PermMap class does not define any new routines beyond those
 *     8-APR-2026 (TIMJ):
 *        Fix astEqual to compare the effective permutation arrays safely when
 *        PermMaps are inverted.
+*     19-JUN-2026 (TIMJ):
```
timj, line 99: @dsberry was the original code ignoring these for a reason?

src/specfluxframe.c
```diff
@@ -1128,6 +1139,31 @@ static int MakeSFMapping( AstSpecFluxFrame *target, AstSpecFluxFrame *result,
    match = 0;
    *map = NULL;
 
+/* If the target and result SpecFluxFrames have equal spectral components
```
timj, line 1142: @dsberry It seemed wrong to me that creating two frames with no arguments did not compare equal but I have no idea if this was an oversight or deliberate.

src/specmap.c
```diff
@@ -1761,11 +1767,11 @@ static double GalVel( double ra, double dec, FrameDef *def, int *status ) {
 
 /* Get the component away from the source, of the velocity of the sun
    relative to the dynamic LSR (in km/s). */
-   s1 = (double) palRvlsrd( (float) ra, (float) dec );
+   s1 = palRvlsrd( ra, dec );
```
timj, line 1770: Definitely the right fix.

src/timemap.c
```diff
@@ -104,6 +104,18 @@ f     - AST_TIMEADD: Add a time coordinate conversion to an TimeMap
 *        - Fix bug in MapMerge that prevented adjacent TAITOUTC and UTCTOTAI
 *        conversions cancelling out.
 *        - Add DTAI argument for TTTOTDB and TDBTOTT.
+*     19-JUN-2026 (TIMJ):
+*        Fix astDat inverse (TAI->UTC) branch for the pre-1972
```
timj, line 108: Two bugs here and they were validated against astropy.

timj (commented):
So a couple of changes here that I'm not sure were deliberate choices for AST but the others seem obviously correct.

src/specfluxframe.c
dsberry: Pretty sure it was an oversight.

src/permmap.c
dsberry: Not that I can see.

PR #59, opened 2026-06-20
fitschan: Do not lose precision when setting RefRA/RefDec values

dsberry:
Fix MakeFitsFrameSet. Previously, precision was lost when copying the new reflon and reflat values into the SpecFrame due to the values being formatted with the default number of digits (7) and then unformatted. Now, the Digits value is temporarily increased to 20 before formatting these values.

This is intended to address the point noted at the end of #58.


dsberry:
These test failures are probably not failures at all but will probably be due to small changes in the created FrameSets caused by increasing the precision on RefRA and RefDec. I'd say the new, More precise, values will be the *right* values. The PR included a fix for such a failure in one of the old ast_tester tests.

timj:
Thanks. I can fix the fixtures.

ast_tester/class-write.fits-class
```diff
@@ -13,7 +13,7 @@ CRVAL3  =               62.938 / Value at ref. pixel on axis 3
 CRPIX1  =           -2820642.0 / Reference pixel on axis 1                      
 CRPIX2  =                  1.0 / Reference pixel on axis 2                      
 CRPIX3  =                  1.0 / Reference pixel on axis 3                      
-CDELT1  =              39062.5 / Pixel size                                     
+CDELT1  =              39062.5 / Frequency resolution (Hz)
```
timj, line 16: I am somewhat baffled by the change in FITS comment here (and also we probably should put the units in square brackets at the start).

timj:
I have regenerated the FITS headers so hopefully the tests pass.

ast_tester/class-write.fits-class
timj: Is it because with this correction the internal (before the special FITS double formatting triggers) value reports no difference from the input header value and so the comment field is inherited from the .head file whereas if AST thinks the value is being recalculated it inserts its own comment?

PR #61, opened 2026-06-23
Add a test for SimpVertices and great circles

timj:
With `SimpVertices=1` (the default) you can end up with polygons being linearized in pixel space. You need `SimpVertices=0` to retain the sky curvature. There was no test for this so here we add a CAR projection test and check that the correct number of pixels are masked out.


timj:
@embray Thanks for the feedback on coverage. I probably should have raised an issue about this first. I started by trying to configure coverage to not report on all the `stopit()` checks that by definition we should never cover and then thought it would be easier to only run coverage on `src/`. I agree that an anomalously low coverage in a test file is a useful thing to know (and probably would have made me notice the region test problem earlier).

I am not wedded to only doing the src/ tree and I'm happy to revisit the ast_tester coverage. Knowing src coverage on its own seems useful but in theory if we can correctly hide the stopit() coverage failures we should be in a position where test code is >95% covered and it will still be obvious when src/ has low coverage.

PR #62, opened 2026-06-28
Ponder a native JSON serialization

timj:
This would also make native YAML pretty straightforward.

This is purely a spec at the moment. No implementation to look at.


dsberry:
Odd that the design doc never mentions FitsChan, only XmlChan and YamlChan. FitsChan does include a native encoding. 

None of the native encodings have ever been used to the best of my knowledge (except obviously for the basic Channel class). I suppose the reason is that the only bit of software that understands the semantics of an AST object is the AST library itself, which already has the Channel class.

timj:
We did use FITS Native form in `ndf2fits` in the past because we knew that GAIA could read it. I think we had ongoing debates with the DS9 people about it since I think you mentioned that they are not using AST in a way that lets AST discover the FITS WCS encoding and so it can't work with native.

We are using the native string form internally in Rubin data but if we could put both the FITS-WCS approximation and the full Native form in the same FITS header and have DS9 pick out the native form in preference, that might be an interesting approach for some users.

The JSON form proposed here is me wondering whether it would be more straightforward for us to write JSON into our JSON models directly instead of a blob, and then have schema validation for the entire model.

cc/ @TallJimbo

PR #63, opened 2026-06-28
Add some reference transformations so we can detect drift

timj:
This was motivated by @embray 's #51 so we can tell if something catastrophically breaks a transform. The FITS versions also test inverses (which led to me realizing that ChebyMap hasn't got the inverse code yet).


embray:
@timj Thanks for this.  Lots going on here; I'll try to look it over later.

ast_tester/transform_oracle_util.c
embray, line 1: Some of these utilities would probably be useful within the other tests as well, `oracle_within_tol` and `oracle_within_tol_wrap` are very useful and probably have equivalents scattered throughout, though I haven't looked.  The Halton sampling also looks useful; I think there are places I might use it.

There is a lot of other boilerplate across the tests, and inconsistencies in how the test files are constructed (a lot inherited from the direct port of the old FORTRAN tests).  Obviously not for here, but a thorough test cleanup on these points is something to consider for the future.

docs/superpowers/specs/2026-06-27-transform-oracle-design.md
```diff
@@ -0,0 +1,414 @@
...
+  tolerance.
+- Give identical pass/fail verdicts on ARM and x86; no CPU-architecture lock-in.
+- Cross-check that the unsimplified (`.map`) and simplified (`.simp`) forms of a
+  fixture produce the same transform output, validating that simplification
+  preserves the math.
+
+## Non-goals
+
+- Bit-for-bit reproducibility of transform output.
+- Performance measurement; that is the job of `perf/bench_simd.c` in PR #51.
```
embray, line 44: ... which might be worth also extending into a general benchmarking program, not just for the SIMD results, again future work.  Maybe integration with airspeed velocity?  Just idle thoughts...

docs/superpowers/specs/2026-06-27-transform-oracle-design.md
```diff
@@ -0,0 +1,414 @@
...
+classes (`Box`, `Interval`, `NullRegion`, `PointList`) which also have a
+well-defined transform.
+
+### FITS-header corpus
+
+`ast_tester/*.head`, 132 files.
+Each file is a set of FITS header cards loaded via an `astFitsChan` into a
+`FrameSet`.
+The mapping tested is `astGetMapping( fs, AST__BASE, AST__CURRENT )`.
+These exercise WCS pipelines, including the sky projections that drive SphMap.
```
embray, line 73: There is now also a growing ASDF corpus (and I have been working slowly on adding more tests for YamlChan).  Maybe a worthwhile future add-on?

docs/superpowers/specs/2026-06-27-transform-oracle-design.md
```diff
@@ -0,0 +1,414 @@
...
+Three knobs:
+
+- Golden checks (A and B) and round-trip: starting defaults `rtol = 1e-12`,
+  `atol = 1e-12`.
+- Equivalence check (C): looser starting defaults `equiv_rtol = 1e-9`,
+  `equiv_atol = 1e-9`.
+  Check C compares two genuinely different arithmetic sequences, a compound chain
+  versus one collapsed operation, which can legitimately differ by more than a
+  few ULP from accumulated rounding, so it needs a looser bound than the
+  same-path golden checks.
```
embray: Makes sense.  Though it might be worth including this explanation in `transform_oracle.h`.  The long-winded comment there does not clarify why there are two different tolerance levels defined (`ORACLE_DEF_TOL` vs `ORACLE_DEF_EQUIV_TOL`).  But also the former is given as a *higher* tolerance now, `1e-7`, than the documented `1e-12` here.  What happened there?  Did it turn out to be impractical?  But now I'm confused.

docs/superpowers/specs/2026-06-27-transform-oracle-design.md
```diff
@@ -0,0 +1,414 @@
...
+
+## Open implementation parameters
+
+These are deliberately left for the implementation phase, with the defaults above
+as starting points:
+
+- The specific low-discrepancy sequence and the number of points per fixture.
+- The base sampling range for native-dump mappings and the clustering criterion
+  and attempt cap for adaptive broadening.
+- The final tolerance values, tuned from the observed corpus spread.
```
embray, line 301: Fair enough.

docs/superpowers/specs/2026-06-27-transform-oracle-design.md
```diff
@@ -0,0 +1,414 @@
...
+- Golden forward/inverse: `rtol = atol = 1e-7`.
+  Golden compares the live output against the committed reference, which may
+  have been generated on a different architecture.  Cross-architecture
+  `libm` differences are far larger than the ~2e-12 seen merely across
+  compilers on one machine: ARM vs x86 CI showed ~5e-9 for a relativistic
+  spectral inverse sampled near its `beta ~ 1` singularity.  `1e-7` absorbs
+  that while still catching real algorithmic change (`>= ~1e-6`).
+- Equivalence: `rtol = atol = 1e-9` (a same-build comparison of two live
+  transforms, so not subject to cross-architecture drift).
+- Round-trip accuracy: absolute `0.05` px (`rtol = 1e-6`).
```
embray, line 418: This makes me wonder if it wouldn't make sense to have per-architecture oracle files, or at some comment in the headers of the `.oracle` files providing provenance (machine architecture and compiler, CFLAGS) that `gen_transform_oracle` (and AST itself) were compiled with.  That might be useful longer-term for meta-comparison across compilers and machines.

ast_tester/gen_transform_oracle.c
```diff
@@ -0,0 +1,322 @@
...
+            oracle_format_double( buf, sizeof buf, in[a][p] );
+            fprintf( fp, "%s%s", a ? " " : "  ", buf );
+        }
+        fprintf( fp, "  " );
+        for ( int a = 0; a < out_n; a++ ) {
+            oracle_format_double( buf, sizeof buf, out[a][p] );
+            fprintf( fp, "%s%s", a ? " " : "", buf );
+        }
+        fprintf( fp, "\n" );
+    }
```
embray, line 179: Maybe nice for readability's sake if these were formatted into fixed-width columns, at the cost of a little extra file size.

ast_tester/gen_transform_oracle.c
```diff
@@ -0,0 +1,322 @@
...
+    const char *root          = argv[1];
+    const char *simplify_out  = argv[2];
+    const char *headers_out   = argv[3];
+    const char *framesets_out = argv[4];
+    astWatch( status );
+
+    const char *hdr =
+        "# transform oracle - regenerate with gen_transform_oracle\n"
+        "# tol: rtol=1e-12 atol=1e-12  equiv_rtol=1e-9 equiv_atol=1e-9\n\n";
```
embray: This being hard-coded is clearly misleading now -- hence some of my confusion about the tolerances.

ast_tester/gen_transform_oracle.c
```diff
@@ -0,0 +1,322 @@
...
+    if ( !ff ) { fprintf( stderr, "cannot write %s\n", framesets_out ); return 1; }
+    fputs( hdr, ff );
+    int nast = 0;
+    char **asts = scan_dir( root, "", "", ".ast", &nast );
+    for ( int i = 0; i < nast; i++ ) {
+        nwritten += emit_fixture( ff, root, asts[i] );
+        free( asts[i] );
+    }
+    free( asts );
+    fclose( ff );
```
embray, line 363: Maybe worth a refactor later since there's a lot of duplication in the corpus generations, maybe latter if when adding additional ones (such as ASDF).

embray (commented):
Impressive.

docs/superpowers/specs/2026-06-27-transform-oracle-design.md
timj: Yes, I'm slowly trying to get code coverage from test fixtures as a source of truth. At some point I need to ponder whether There should be an explicit serialized FrameSet per FITS header so that there is a direct means of seeing what we expect a header to generate. Sometimes this is encoded in the FITS output being Native but it's all a bit inconsistent. Some of the new mapping fixtures I just added came from dumping the frameset out after the FitsChan read. There does need to be some tidying up of test fixtures.

docs/superpowers/specs/2026-06-27-transform-oracle-design.md
timj: Getting Claude to go back and fix up design docs after the fact is easy to forget.

docs/superpowers/specs/2026-06-27-transform-oracle-design.md
timj: I was wondering whether it was better to generate per-architecture oracles or have a bigger tolerance. On the one hand if arm generates wildly different values from x86_64 we really do need to know about it since that should be investigated. On the other hand if you know that you should get the exact x86_64 values every time you run it then if you get a slightly different value and you didn't change anything that might be telling you something important.

The only "safe" option is to generate for both and compare exact variants with the same arch but also validate the architectures to make sure they are not wildly different.  Not sure the best way to do that. I suppose we could store a diff for each transform relative to x86.

ast_tester/gen_transform_oracle.c
timj: Not entirely sure if we have predictable column sizes but can take a look. I will need to regenerate anyhow now that I added more than a hundred new simplify fixtures.

embray: Just a minor thing really.  I thought maybe because there's a specific function for outputting fixed-width floats.

docs/superpowers/specs/2026-06-27-transform-oracle-design.md
timj: A benchmark would be useful. Will add it to the list.

ast_tester/transform_oracle_util.c
timj: I also think that we can probably improve the diff comparator for the serializations to spot if the line diff is only a float and then do a float comparison instead of string comparison.

src/polymap.c
```diff
@@ -151,6 +151,12 @@ f     - AST_POLYTRAN: Fit a PolyMap inverse or forward transformation
 *     28-JUL-2025 (DSB):
 *        - Add protected virtual method astMergeShift.
 *        - Add public static method astShowPoly.
+*     2-JUL-2026 (TIMJ):
+*        The default for IterInverse is now zero if the PolyMap has
```
timj, line 155: @dsberry forcing lots of transforms and inverse transforms through every single mapping type led to a couple of anomalies in polymap, slamap, and splinemap. Can you please take a look and see if they make sense to you? It's trying to ensure that BAD is returned for a mapping instead of raising an error or returning NaN.

dsberry: Yes, those fixes all look good to me.

ast_tester/transform_oracle_util.c
embray: Could pick that from https://github.com/Starlink/ast/pull/37/changes#diff-279af37bcf88a56a1895f24837a12d2b95cbb35b4ab5ea77fc4d1ea8a4eca6b0 , though if you prefer not have a Python dependency it could be rewritten in C easily enough I think.

PR #66, opened 2026-07-03
Miscellaneous YamlChan bug fixes and test coverage

embray:
This is built on top of #55 (otherwise ready to merge). I got kind of carried away with several things:

- Reworks on top of #55 to use the new Object-attached KeyMap to store additional data about transforms read from ASDF
- This is used especially for the `divide` transform to simplify round-tripping it from/to YAML without going through the awkward `FindDivide` code path.
- Simplified how the `spherical_cartesian` transform is handled; my original implementation overcomplicated things and was at odds with out other transforms handled angular unit conversions (which I didn't fully understand at the time I implemented it).
- Added several more tests for YamlChan.  My initial goal was to try to get to 80% coverage, but for now will stop at ~76%.  There are a few more cases that could be hit fairly easily but the returns in terms of line count are rapidly diminishing.
- Fixed a few small bugs caught thanks to the added tests.


src/yamlchan.c
```diff
@@ -106,6 +106,19 @@ f     The YamlChan class does not define any new routines beyond those
...
+*        unit conversions when simplifying the transform chain.
+*     2-JUL-2026 (EMB):
+*        - Fix reading of the ASDF linear1d transform, which used the wrong
+*        property name and constructed an incorrect Mapping.
+*        - Fix reading and writing of observer earth locations, which failed
+*        to recognise the earthlocation class and wrote scalar quantities in a
+*        form that could not be read back.
+*        - Fix serialisation of a YamlChan object itself: it was missing from
+*        the AST object loader and had an out-of-bounds array access when
+*        dumping a channel that uses the NATIVE encoding.
```
embray, line 124: Stupid question about the History entries--as I originally worked on this way back in July, it appears, that's when the history entries are dated to.  But this PR has stalled, and there has been a release since I last worked on it.  So would it make sense to bump the dates on the history entries, e.g. to today?  Or does it matter?

timj: The History is one of those things where the Starlink policy was defined in a world in the 1980s where source control was not really used. Obviously today you would look in the git history but it seems like we should not break with precedent. My feeling is that you can leave it or you can merge both records and summarize with a new date -- that seems consistent with the actual requirement.  For #81 the history record is ridiculous and needs tidying because there are multiple records covering a couple of days which should clearly be turned into a single record.

embray: Given that the precedent exists anyways I think it's not a bad one.  Sometimes a high-level summary, like a changelog, is a lot easier to read than the full version control history.  I think I'll merge/summarize with new date like you suggested.

PR #67, opened 2026-07-07
yamlchan: Add libfyaml support as an alternative to libyaml

embray:
Something I've been working on off-and-on for a few weeks and finally had a chance to finish.

## Motivation

AST uses the older (but still actively maintained) lower-level `libyaml` to parse and emit YAML.  My project, `libasdf`, on the other hand uses the newer `libfyaml`.  And then, in particular, the [libasdf-gwcs](https://github.com/asdf-format/libasdf-gwcs) plugin for `libasdf` which is using AST as an evaluation backend, in turn also requires `libfyaml` *and* via AST `libyaml`.

In effect, any code that depends on `libasdf-gwcs` must also detect two separate YAML libraries, which is a bit onerous.  It's not the end of the world--both libraries are widely available.  It just feels a bit rough.

On the other hand I'd also understand if AST doesn't want to maintain support for both.  I just wanted to see if it *could* be done, since both libraries are implementing the same recommended YAML parser/emitter interfaces, just with slight API differences.

## Implementation

This adds a `yaml_backend.h` header-only library providing `astYaml` functions that roughly map to the equivalent `yaml_` functions originally used by `yamlchan.c`.  This is an almost direct drop-in replacement, so the changes to `yamlchan.c` itself are minimal; the only major difference is in error handling which is improved over the original implementation anyways (better reporting of line number context for errors).

Between the two backends the `libyaml` one is still much simpler--it's almost a direct wrapper.  The `libfyaml` backend needs to do a little extra work to adhere to the original `libyaml` interface.

The backend to use can be selected explicitly when configuring the build.  If not specified the build system chooses, with priority given for `libyaml` for backwards-compatibility: The outputs of the two libraries are not bytewise-identical, though they are of course equivalent YAML.

## Shortcomings

Compile-time backend selection: The YAML backend (libyaml, libfyaml, or none) is selected at CMake/autoconf configure time and compiled into libast. Packagers who want to ship both variants must build separate conflicting packages (there is precedent for this, however.). A cleaner long-term approach might be a satellite shared library, analogous to the existing libast_grf_*.so graphics plugins.  Both backends could co-exist with one selected and `dlopen`'d at runtime, though this would be a more complex refactor.


timj:
Since the YAML handling is mostly motivated by the ASDF needs I don't have a problem with adding support for a new YAML library. If I was a packager I don't really understand why I would have to build both variants -- wouldn't I always choose fyaml?  There is a conda-forge package of AST which would need tweaking if we switched YAML library. We picked libyaml originally because we didn't really know any better.

embray:
> wouldn't I always choose fyaml?

Possibly.  Right now the only reason to do so would be my specific use-case really, so it has to be an informed choice.

> There is a conda-forge package of AST which would need tweaking if we switched YAML library. We picked libyaml originally because we didn't really know any better.

I'd be more fair to yourself than that :)  At the time libfyaml was quite new and not at all mature yet (its first release on github was only in 2019).

Depending on how dependent upstream users are on byte-for-byte identical outputs from AST, it might still matter, as there are cosmetic differences in what the two libraries output.  I actually find that some of libfyaml's default output formatting is a bit ugly (compromises it makes, I think, as a performance tradeoff).

For example, even short flow-style sequences like `[a, b, c]` are not collapsed onto a single line; instead it outputs:

```
[
  a,
  b,
  c
]
```

which I find a bit annoying.  Of course, for the event stream interface doing this requires some buffering/look-ahead of events. 

Those cosmetic output differences are my only concern though.

I also plan when I have time to propose some fixes to libfyaml for these formatting issues.  I think some options to tweak this might be acceptable, where performance is less of a concern--its maintainer has been amenable to some of my past suggestions.

embray:
The ASAN build is revealing a number of memory leaks.  Strangely some of them don't seem directly related to this branch, while others do, which suggests to me that the correct flags aren't being passed to ASAN to cause tests to halt, but I'll have to have a look at that.

At least one of the leaks is a bug in libfyaml itself, fixed in v0.9, though Ubuntu <= 24.04 has a version of libfyaml with this bug and it doesn't appear to be patched.  I'll keep investigating but that one might not be anything that can be easily worked around and should maybe just be documented.  It's a small leak that only occurs while writing.

timj:
Interesting. As I have expanded test coverage I have found some ASAN issues. Sometimes it's the system not including an astBegin/astEnd in the right place but others have been real. So you are saying that some of these are in general mapping code?

embray:
> So you are saying that some of these are in general mapping code?

No, just some specific to `yamlchan.c`.  One that I introduced in #39 which I'm fixing now, and a couple unique to this PR, also being fixed.  There is one leak caused by a bug in libfyaml mentioned above.  I'm not sure I can work around it so I might just make an exception for it (it's fixed in newer versions).

embray:
Fixed all leaks exposed in `testyamlchan` that I could find--some pre-existing and some introduced originally by this PR.  For the one upstream bug I couldn't workaround, so I added suppression file: https://github.com/embray/ast/blob/88b2aa90ea1e428455fb4b5e056f260d0c84a402/ast_tester/libfyaml.supp

```
ASAN_OPTIONS=detect_leaks=1;LSAN_OPTIONS=suppressions=../../ast_tester/libfyaml.supp ./testyamlchan
  parser error message (expected):
AST: Error in routine test_parser_error at line 858 in file /home/embray/src/asdf-format/starlink/ast/ast_tester/testyamlchan.c.
astRead(YamlChan): libfyaml parse error:
libfyaml parser: flow sequence without a closing bracket at line 4, column 1
 All YamlChan tests passed (libfyaml)
-----------------------------------------------------
Suppressions used:
  count      bytes template
    183       4487 fy_emit_event_vcreate
-----------------------------------------------------
```

(the error in the output comes from one of the tests and is expected)

PR #68, opened 2026-08-09
More fixes for bugs found by Claude

timj:
@dsberry  this is now ready for review. A bunch of fixes Claude find in KeyMap and from analyzing round trip semantics in serialization. There is one open question at the end regarding IsSimp placement being incorrect due to objects changing state when they should not be. I would really like your opinion on that. The suggestion in the doc describing the issue is to add an extra state in Mapping. This is solving a problem where it's not possible to reason where the `IsSimp` flag will end up in the serialized output without knowing deep internals that are not predictive.

From Claude:

# Defect fixes in KeyMap, FitsChan and the rounding idiom

A set of correctness fixes to `libast`. Nineteen defects are fixed, one commit each, spanning `src/keymap.c`, `src/fitschan.c` and a rounding idiom used across the library. Regression tests and any forced reference fixture updates are included alongside the relevant fix. One further defect is investigated and documented but deliberately left unchanged pending maintainer input (see the end of this document).

39 commits. 17 files under `src/` (+669/−181), 71 regenerated reference files, and new test coverage in `ast_tester/testkeymap.c` (+684) and `ast_tester/testfitschan.c` (+60).

All work was validated with both a Release build and the CMake developer build (`-DAST_C_STANDARD=11 -DAST_ENABLE_WARNINGS=ON -DAST_ENABLE_SANITIZERS=ON`). The full suite (1039 tests) passes in both. The changes add no new warnings: per-file warning sets are byte-identical to the merge base for every modified source file.

## Rounding

The idiom `(int)( x + 0.5 )` truncates toward zero, so it rounds correctly only for non-negative operands. Three sites had genuinely negative operands and were wrong; the remainder are converted so that one rounding idiom now remains in the library instead of two.

### 1. Negative grism interference orders were destroyed
**File:** `src/fitschan.c` (`GrismSpecWcs`)

The interference order was read from FITS keyword `PVi_1` with `(int)( pv + 0.5 )`. Negative diffraction orders are physically routine: an order of −1 was rounded to 0, which is degenerate, so `astRead` rejected an otherwise valid header with "unusable parameter values"; an order of −2 was stored as −1, producing a WCS with the wrong dispersion and no diagnostic at all. Positive orders are unaffected. Regression test added.

### 2. Every negative value read as an integer was off by one
**File:** `src/keymap.c` (`ConvertValue`)

A stored Double, Float or numeric String was converted to an integer type with `(int)( dval + 0.5 )`, so every negative value was rounded toward zero by one, exact integers included: `astMapGet0I` on an entry holding −2.0 returned −1. The existing coverage used 1999.9, a positive value that rounds identically either way, which is why the defect went undetected. New checks cover an exact negative integer, values either side of a tie, and a positive control.

### 3. `WCSAXES` was truncated rather than rounded
**File:** `src/fitschan.c` (three sites)

Each site read `wcsaxes = (int) dval + 0.5;`. The cast binds tighter than the addition, so `dval` was truncated to `int`, `0.5` added to the *integer*, and the result truncated again on assignment — the `+ 0.5` was dead code. A header carrying `WCSAXES = 1.9999999999` failed `astRead` outright with a spurious "longitude axis without a corresponding latitude axis", because the axis loop ran only once. No `(int) X + 0.5` construct now remains anywhere in `src/*.c`.

### 4. An axis title was drawn one pixel out of position
**File:** `src/grf_log.c` (`SvgX`, `SvgY`)

`SvgX`/`SvgY` convert a world coordinate to an SVG pixel coordinate for anything drawn in the virtual viewport, including axis titles, so their input can genuinely be negative — unlike the other sites in this sweep. Truncation placed the "Right ascension" axis title one pixel right of its correct position. `ast_tester/serpens.svg` is regenerated by running `testgrid` rather than hand-edited; the only change is the label moving from `x="-11"` to `x="-12"`.

### 5. Consistency conversions
**Files:** `src/axis.c`, `src/mapping.c`, `src/mathmap.c`, `src/matrixmap.c`, `src/moc.c`, `src/pcdmap.c`, `src/polymap.c`, `src/wcsmap.c`, `src/winmap.c`, `src/yamlchan.c` and the graphics back ends

Operands at these sites are structurally non-negative — axis indices, counts, polynomial powers, colour indices and the like — so behaviour is unchanged. The value of the change is that a reader need not work out whether a given `+ 0.5` was chosen deliberately or merely inherited.

### 6. `astMapIterate` never terminated when `SortBy` was set
**File:** `src/keymap.c` (`MapIterate`)

`AddToSortedList` builds a circular doubly linked list, so an entry's `snext` pointer is never NULL. `astMapKey` accounts for this and stops when the walk returns to the first entry; the sorted branch of `MapIterate` did not. Iteration signals its end by returning a NULL key, and the key was taken from an entry that is never NULL, so a caller's loop ran forever, re-yielding the same keys in order. The unsorted branch is unchanged: it walks the per-bucket chains, where NULL does mark the end.

### 7. A refused `astMapRename` destroyed the entry
**File:** `src/keymap.c` (`MapRename`)

The entry was removed from the hash table before `MapLocked` was tested. `AddTableEntry`, which puts it back, is reached only while the status is good, so the `AST__BADKEY` branch fell through to `FreeMapEntry` instead. A refused rename left the KeyMap holding neither the old key nor the new one, and the value was lost. Whether a rename would introduce a new key can be determined without mutating anything, so the error is now reported before the entry is touched.

### 8. Out-of-range floating point to integer conversion was undefined
**File:** `src/keymap.c` (`ConvertValue`, new `DtoI`/`DtoS`/`DtoB`/`DtoK` helpers)

A stored Double or Float was cast straight to the requested integer type. C11 6.3.1.4p1 makes that undefined outside the destination's range, so `astMapGet0I` on an entry holding a value too large for `int` returned whatever the toolchain happened to produce, with nothing in the call signalling it. Conversions now saturate at the destination's limits; a negative value saturates at zero for the unsigned type. Every in-range value converts as before.

### 9. NaN input to the conversion helpers was undefined
**File:** `src/keymap.c` (`DtoI`, `DtoS`, `DtoB`, `DtoK`)

A NaN satisfies neither saturation comparison and so fell through to the cast itself, which the same C11 clause leaves undefined. The helpers now return 0: a NaN has no integer value, so any result is arbitrary and the only requirement is
that it be defined.

### 10. Numeric string overflow was silently wrong
**File:** `src/keymap.c` (`ConvertValue`)

The String-to-integer arms scanned with `" %d %n"` and accepted the result when the whole string had been consumed. C11 7.21.6.2p10 makes a `sscanf` numeric conversion undefined out of range, and glibc does not fail — it wraps the accumulated value and reports success. The acceptance test therefore passed and the existing `"%lf"` fallback was never reached, so `"9999999999"` converted to 1410065407. `strtol` with an `errno` range check is used instead, so out-of-range strings fall through to the `"%lf"` attempt exactly as any other unparsable string does.

### 11. A zero-length vector produced an unreadable entry, and could crash `astWrite`
**File:** `src/keymap.c` (`MAKE_MAPPUT1`, `DumpEntry`)

A KeyMap entry records a vector length of zero to mean *scalar*, so there is no representation for a zero-length vector. The put macros applied no check to the supplied size, because the per-type check macros are empty for the numeric types, so a size of zero created an entry claiming to be a scalar whose value pointer was null. `DumpEntry` then took the scalar branch and read that pointer as the payload: the Object type had a guard, the string type did not, and `astWrite` dereferenced null. The call is now rejected, and the missing guard added to the string branch so the dump path cannot dereference a null pointer.

### 12. `astMapGetElem<X>` reported success for an undefined entry
**File:** `src/keymap.c` (`MAKE_MAPGETELEM`, `MapGetElemC`)

The result was set to one as soon as the entry was found. For an undefined entry the raw value pointer is null and the conversion step is skipped, so the function returned success without writing anything to the caller's output buffer — a caller trusting the return value read whatever was already there. `astMapGet0<X>` and `astMapGet1<X>` both report failure for the same entry, so three functions answering one question gave two different answers. An out-of-range index still reports `AST__MPVIN`, because the bounds check continues to run first.

### 13. No freshly written KeyMap matched its own re-dump
**File:** `src/keymap.c` (`astLoadKeyMap`)

`KeyCase` is the one KeyMap attribute whose unset value is −1; `KeyError`, `MapLocked` and `SortBy` all use `-INT_MAX`. The `KyCas` card was read with a default of `-INT_MAX`, which is not equal to −1, so `TestKeyCase` reported an absent attribute as explicitly set and the next dump wrote it out. The sentinel itself is unchanged, since altering it would change the attribute's public default value expression.

### 14. A locked, non-empty KeyMap could not be read back
**File:** `src/keymap.c` (`astLoadKeyMap`)

Every attribute, `MapLocked` included, was applied before the entry-reading loop. A KeyMap is empty at that point, so every key the loop then put was a new key, and the shared guard in the put macros refused it with `AST__BADKEY`. `astWrite` succeeded and produced an `MpLck` card, but the matching `astRead` failed and returned a null object. Reading the card after the entries also propagates the flag into nested KeyMaps held as entries, which the earlier placement could not do because those entries did not yet exist.

### 15. A per-entry `mem%d` read destroyed `SortBy=Age*` ordering across a load
**File:** `src/keymap.c` (`astLoadKeyMap`)

A `mem%d` card was read for each entry and used to seed the member counter before adding that entry. `DumpEntry` writes no such card, so the default of zero was used on every iteration: the counter was reset before each put and every loaded entry was given member number zero. Since those numbers determine `SortBy=AgeUp`/`AgeDown` ordering, the read actively destroyed that ordering across a load rather than merely failing to preserve it. Entries now receive sequential numbers in the order they are read.

### 16. Copied keys used the source KeyMap's casing rule
**File:** `src/keymap.c` (`MapCopyEntry`, `MapCopy`, `CopyMapEntry`)

`MapCopyEntry` converted the caller's key once, using the *source's* `KeyCase`, and used that one string for both the source and destination lookups — so the destination was searched under a rule that is not its own and could miss a key its own rule would have matched. `MapCopy` hashed each source key into the destination table without converting it at all. The stored key was wrong as well: `CopyMapEntry` copies the key text verbatim, so a copied entry was filed under the source's casing. Since `KeyCase` can only be changed while a KeyMap is empty, every other route into a KeyMap stores a key already expressed in that KeyMap's own case; the copy paths were the sole exception.

### 17. Dumps never converged
**File:** `src/keymap.c` (`Dump`)

`Dump` walked the hash table bucket by bucket and followed each chain from the head. `astLoadKeyMap` re-puts entries in the order it reads them, and `AddTableEntry` inserts at the head of the bucket, so any two keys sharing a bucket swapped places on every write and read cycle and the dump never converged. Whether two logically identical KeyMaps serialised identically depended on their insertion history and on how many times each had been round-tripped. Entries are now emitted in key order, which is idempotent under any insertion order. `Dump` already ignored `SortBy`, so no documented behaviour changes; the hash table and the `MapSz` card that records its size are unaffected.

### 18. Two hand-written type variants missed the fixes their siblings received
**File:** `src/keymap.c` (`astMapPut1AId_`, `astMapGetElemAId_`)

These two are written out longhand rather than generated from the macros that cover the other types, so they did not receive the guards added in items 11 and 12. `astMapPut1A` with `size=0` silently created an entry with `nel == 0` that is read back as a *scalar* with a NULL value, for which `astMapGet0A`, `astMapGet1A` and `astMapGetElemA` all returned success; `astMapGetElemA` on an undefined entry returned 1 with the caller's buffer untouched, handing back an uninitialised `AstObject *`. Both are reachable from the public C API and from Fortran. `MapGetElemC`, a third hand-written variant, already had the guard.

### 19. Wide float values were written one column too far right
**File:** `src/fitschan.c` (`EncodeFloat`)

A redundant leading zero in an exponent is removed by sliding everything to its left one place right and writing a space into the vacated first character — the shift is what keeps the value right-justified, and it works whenever there is padding to consume. When the two-digit exponent form already filled or exceeded the field width there was no padding, so the space survived: the value was written starting one column further right than every other value in the header, and the field ran one character past its intended end. A leading pad space is now consumed when one exists; otherwise the gap is closed and the string shortens, matching the idiom already used when inserting a decimal point.

### 20. `WcsNative` serialised an invert flag acquired after encapsulation
**File:** `src/fitschan.c` (`WcsNative`)

One PermMap was built and used twice: forward for the stage 1 CmpMap, then inverted for stage 3. `astCmpMap` clones rather than copies its components, so the stage 1 CmpMap held the same object that was subsequently inverted. It recorded an invert flag of zero at construction and so dumped no `InvA` card, yet its child PermMap dumped `Invert = 1` — a state acquired only after being encapsulated. The flag never reached the arithmetic, because `Transform` forces the parent's recorded value onto the child, but it was serialised, and a reader trusting it would need the parent's `InvA` to cancel it back out. Stage 3 now uses a copy.

### 21. Legacy `CDjjjiii` cards were dropped under encodings that ignore them
**File:** `src/fitschan.c` (`SpecTrans`)

Every element of the legacy six-digit `CDjjjiii` matrix was read unconditionally, but the result was stored — as `PCj_i` plus a unit `CDELTj` — only under the FITS-IRAF encoding. Reading a card marks it used, so a header carrying `CDjjjiii` lost those cards under every encoding while only FITS-IRAF gained the rotation matrix they describe. Under the others the values were read, discarded, and the cards marked used, so they were dropped from the output entirely instead of being either honoured or passed through unchanged. The cards are now read only under the encoding that uses them, and survive untranslated elsewhere.

### 22. A repeated keyword was only half-consumed
**File:** `src/fitschan.c` (`GetValue2`, new `MarkAllCards` helper)

Reading a FITS-WCS header consumes cards through two paths that treated repeated keywords differently. `WcsFcRead` sweeps every card and marks each match, so every copy is consumed. `SpecTrans` reads through `GetValue`, a first-match forward search, so it marked exactly one card however many copies existed. For a header holding two HDUs' worth of WCS cards, a single `astRead` therefore stripped both copies of `CRPIXi`, `CRVALi`, `CTYPEi` and `EQUINOX` but left the second copy of `CDi_j`, `RADECSYS` and `DATE-OBS` behind. What survived was a partial, self-inconsistent WCS description, and which keywords survived was decided by which read path happened to handle them.

## Reference file changes

69 `.native` files, one `.fits-wcs` and one `.svg` are regenerated. The `.native` churn is the `EncodeFloat` justification fix (item 19) and the repeated-keyword fix (item 22); `serpens.svg` is the one-pixel axis-title correction (item 4). Each was regenerated by running the relevant test rather than hand-edited, and reviewed against the change that caused it.

## Worth attention on review

- **`SortBy=Age*` no longer survives a dump/load.** Two individually correct changes compose to it: entries are dumped in key order (item 17) and load assigns sequential member numbers (item 15). The dump writes only `MemCnt` and no per-entry member card, so Age order is unrecoverable by design. Probing showed the merge base only *appeared* to preserve it — 3 of 4 trial key sets failed there too — so this hardens an already unreliable behaviour into a reliably absent one rather than removing a working feature. Worth a line in the `SortBy` documentation saying Age ordering does not survive serialization.

- **`Strict=1` now rejects dumps written before item 15**, failing with `The value "mem1 = 0" was not recognised as valid input`. A deliberate consequence of no longer reading a card that nothing writes, but it is a compatibility break.

- **Item 22 covers the `SpecTrans` path only.** `DSSToStore` and `SAOTrans` still consume just the first copy of a repeated keyword, so the asymmetry persists for DSS and SAO headers. Deliberate scoping, but narrower than the commit message implies.

- **`MemCnt` is inconsistent with the loaded entries.** `Dump` writes the total member count and load then numbers entries `0..n−1`, so a reloaded object's `member_count` does not correspond to its entries' member numbers. Harmless today because nothing reads both, but it is an invariant violation sitting in a serialization path.

- **Integer-to-narrower-integer narrowing is still unfixed** while float-to-integer now saturates (item 8), so the same numeric value converts differently depending only on its storage type: an `int64` holding 2^40 read as `int` wraps to 0 where a `double` saturates at `INT_MAX`. The merge base behaves identically and this is a separate defect class from the one scoped here, so it is left for a follow-up rather than widened into this branch.

- **The test suite is weak evidence for serialization changes.** The only byte-comparison path for FITS card text is `add_wcsconv_test`, and every `native` row in `ast_tester/wcsconv_tests.txt` sets `skip_string_compare=yes` — so only the `_astequal` variant runs, and it compares re-parsed objects numerically. A green suite does not gate card consumption or byte-level output. The new KeyMap byte-stability test was added precisely because nothing gated item 17: it asserts round-trip idempotence *and* insertion-order independence, on keys chosen to collide in a single hash bucket, since that is the only configuration where the property is non-trivial.

---

## Related: `IsSimple` tag placement depends on aliasing history (documented, not changed)

**File:** `docs/issues/2026-08-09-issimple-tag-placement.md` — new document, no source change.

`AST__ISSIMPLE_FLAG` is a single bit of `Mapping.flags` carrying two unrelated meanings: "do not simplify this again" and "write `IsSimp = 1` when dumping". `astSetInvert` clears it unconditionally, including when the value being set is the value already stored. Routines with a documented read-only contract — `astEqual` among them — decompose a Mapping with `astMapList`, which returns reference-counted pointers to the *live* components, flip a component's Invert flag to line it up for inspection, and then restore the *value* but not the bit
the flip incidentally cleared.

Which node in a serialized Mapping tree carries `IsSimp = 1` is therefore decided by which nodes happened to be aliased into some earlier probe, not by the shape of the tree or by which parts of it have actually been simplified. Two Mappings that `astEqual` reports as equal can serialize differently. Three demonstrations using only the public C API are included in the document.

An instrumented run of the full suite recorded 25,644 `astSetInvert` calls that cleared an already-raised bit: 9,458 from 38 routines whose contract is read-only, across 17 files, and 16,186 from 10 routines that change Invert deliberately. The document costs out both candidate repairs — splitting the flag (2 files, ~5 sites, every probe unchanged) against making the probes non-mutating (28 files, 98 routines, 211 save/restore brackets, 60 of those routines unexercised by the suite) — and recommends the split.

The open question for the maintainers is which of the two meanings the public, read-only `IsSimple` attribute should report after a split. Its documentation describes the guard meaning, which argues for keeping it there; against that, reporting the dump bit keeps `astGetI( map, "IsSimple" )` consistent with the `IsSimp` card in a dump of the same object. The attribute has no in-tree consumers outside `mapping.c`, so either choice is safe to implement; it is an API-semantics decision rather than a mechanical one. No change is made pending that decision.


dsberry:
@timj I'm not sure what the major problem is with IsSimple. It was only ever intended as a hint that simplifying the mapping may be a waste of time. If IsSimple==1, then the Mapping has definitely been simplified, if IsSimple==0 (the default) then it may or may not have been simplified.  So dropping an IsSimple==1 setting is not a major problem - it just changes the statement from "definitely" to "may be". It may result in the Mapping being unnecessarily simplified again, but that's not really an error as such.  If the problem is that dumping two supposedly identical Mappings will produce slightly different dumps (one with IsSimp and one without), is that a problem? The test of equality is not whether the text dumps are identical, it's whether the astEqual function returns non-zero. Text dumps of equal objects are often different - for instance, if one includes comments or defaulted attribute and the other doesn't.

So what is the actual fundamental problem that we're trying to fix here?

timj:
I think the fundamental problem is that doing a read-only operation like `astEqual` can flip the IsSimple flag and because some of the mappings are cloned this can change how the output is serialized based on an internal implementation detail on how the tree was scanned when `astEqaul` was called and whether there were aliases or not. This makes it impossible to reason about how the serialized output should look if you are trying to treat the serialization as a reference gold standard.

timj:
@dsberry what did you think of the other fixes?

If you couldn't find the longer description of the simplify issue it's at 
https://github.com/Starlink/ast/blob/c1242102c06096b134337589ca8248480d476a2d/docs/issues/2026-08-09-issimple-tag-placement.md

dsberry:
@timj  I tend to agree with claude that option A is better. In terms of which flag should be returned by astGetC( map, "IsSimple" ), I'd say that sticking to the documented meaning of "No need to re-simplify" is better - the details of what does and does not go into a dump is a purely internal affair (the dump format is not part of the public contract). But you're the boss now :-)

Regarding the other fixes, I had a quick glance through them and didn't spot anything to worry about.

timj:
@dsberry I've had a go at fixing the `astEqual` simplification flag side effect. Ended up adding internal state for whether the inverted or non-inverted mapping had been simplified or not. This does result in some previously simplified mappings now reporting that they have been simplified in the dump. The last commit tries to clear the simplification flag if the mapping itself is mutated -- this seems like the right thing to do and it seemingly found a bug in the way PolyMaps are copied. The problem I then encountered is that in the C when you add mappings to a CmpMap clones are stored in the CmpMap and so if you then edit the original it edits the version in the CmpMap -- so it's possible to modify it in such a way that it would now simplify differently (the code now clears the simplify flag on mutation) but crucially the CmpMap still thinks it has been simplified, presumably meaning that `astSimplify` will not do anything at all.

I'm not entirely sure what the best approach is here. Make it so that CmpMap always tries to re-simplify (so never set IsSimp for a CmpMap)? Even if the mappings were copied (which for all I know will break existing code with people relying on it being a clone) instead of cloned I'm not sure if that would fix things since no-one really wants a read only mapping do they? This is not solely a serialization issue at this point.

timj:
Claude has done some analysis of the ways to handle the stale IsSimp flag in CmpMap in [2ff4581](https://github.com/Starlink/ast/pull/68/commits/2ff458157794670ed1928830e8b99ce1316d9967) -- obviously removing the IsSimp flag on CmpMap fixes things but is then slower every time you call astSimplify. Whether that's a problem depends on how often you are calling simplify. The other point I suppose is whether we know exactly what changes can be made to each mapping that would invalidate the simplification flag. If that is limited maybe there is a compromise solution

Current status of the 3 commits from today (not written by me):

IsSimple record: three fixes and one open question

# 823a33d0 — record the simplified state of each orientation separately

AST__ISSIMPLE_FLAG was cleared unconditionally by astSetInvert. The many routines that set Invert only to line a Mapping up for inspection and then set it back — astEqual among them, via cmpmap's Equal and astMapList, which hand back the live components rather than copies — left the Mapping claiming it needed re-simplifying. Dump writes the IsSimp card from that record, so which nodes of a serialized tree carried the card depended on which of them something had inspected, and two Mappings that astEqual reported as equal could dump differently. No other implementation could predict where the card belongs.

Now recorded per orientation, in ISSIMPLE and ISSIMPLEINV, selected by the current Invert value. Set Invert and set it back and you select the same flag again, so the Mapping reports exactly what it did before; leave it inverted and you select an orientation that was never simplified. astSetInvert clears nothing, and Dump, astLoadMapping, astSimplify_ and GetAttrib are untouched — one functional line changes in mapping.c.

Side effects: IsSimple keeps its documented meaning and once again agrees with the IsSimp card; setting Invert to the value it already holds no longer clears it; astClearInvert on a Mapping simplified while inverted now reports zero, which is correct but differs from before. 27 fixtures gain 67 IsSimp cards — nothing removed or moved — regenerated from unchanged .map inputs.

# 7a0729c5 — discard the record when a Mapping is changed

Invert was the only change that discarded the record, but eleven attributes across five classes alter what a Mapping does or how it simplifies and touched nothing: IntraMap.IntraFlag, MathMap.Seed/SimpFI/SimpIF, PermMap.PermSplit, PolyMap.IterInverse/NiterInverse/TolInverse, WcsMap.FITSProj/LonCheck. SimpFI/SimpIF are the sharpest — they decide whether a forward-inverse MathMap pair cancels. Setter and clearer of each now discard it.

astPolyTran and astMergeShift copy the PolyMap and replace a transformation in the copy, which inherited a record describing coefficients it no longer had. In astMergeShift this was suppressing real work: the copy is handed to astSimplify to amalgamate coefficients on the same powers, which was skipped whenever the input PolyMap had been simplified.

Frames are unaffected throughout — they report zero regardless — which covers Regions and FrameSets and their mutable attributes. No fixture changed.

# 2ff45815 — report on the remaining gap

Containers clone their components, so mutating a component through the caller's own pointer discards the component's record and leaves the container's intact. Demonstrated with the public API: the CmpMap's dump changes and still carries its card. Nothing can propagate the discard upward. Invert is the exception and is safe, because CmpMap, TranMap, RateMap and SwitchMap snapshot each component's Invert and compensate for a later change rather than merely noticing it.

The report costs four repairs and recommends never reporting IsSimple for a Mapping that holds components — withdrawing the claim rather than validating it, and needing no new mechanism since Frame already does exactly this. A change counter is rejected because it records events rather than state, so a save-change-restore would invalidate every enclosing container: the same history-dependence, one level up.

Two measurements back it: only 3.7% of astSimplify calls on a CmpMap short-circuit (against 95% for ShiftMap, 90% for SlaMap), and a timed WCS workload over 132 real headers costs 0.4% overall. The one pattern where the guard earns its keep is asking a deep, already-simplified WCS chain for its Mapping repeatedly — 4.4×, or 27 µs a call — so the recommendation is conditional on no caller doing that at volume.

## Tests and verification

Four new checks in ast_tester/testmapping.c: a dump is unchanged by comparing the Mapping with astEqual; IsSimple and the dump both track Invert across a no-op set, an inversion and a restore; and setting an attribute, clearing an attribute or fitting a replacement transformation each discard the record. Each was watched failing before the fix.

Both build trees pass 1061/1061 — Release, and Debug with sanitizers — with no new compiler warnings.

## Open

Whether to adopt Option D. It is six GetIsSimple overrides, roughly 52 tests of fixture churn, and one assertion in testequaldumpstable moving from two cards to one.

timj:
I'm inclined to leave the fixes as is (they do reflect the correct state of whether something has been simplified or not) and leave the "in theory you can invalidate a simplification flag on a CmpMap by tweaking the mapping in the CmpMap" to a documentation warning. If people are tweaking the thing they added to the CmpMap after they added it they are already risking unexpected changes in the CmpMap itself -- maybe that's a bug anyhow and the CmpMap should be copying.

dsberry:
Sorry, been off air for a while. I think you're right. See "Changing Attributes of a Mapping" in the docs.

David

timj:
@dsberry thank you!

> I think you're right. See "Changing Attributes of a Mapping" in the docs.

This was the clue I needed. It turns out that half the mappings were correctly refusing to change attributes when cloned but many of the newer ones were not doing that. I've now changed it so that all the mappings declare they are immutable after cloning and this all makes a lot more sense. All the ones we were testing (like PolyMap) did not have the check but ZoomMap did.

This obviously means that the fixed code might break some users but it's clearly the intent and they are doing something a bit dangerous and should know about it.

PR #72, opened 2026-08-26
Reorganize test fixtures into a hierarchy

timj:
The ast_tester directory continued to grow test fixtures without any kind of organization. Here I attempt to add some hierarchy and organization.

@embray Does this mess up any of your active branches?

The diff is large but is mostly moving files around. The fortran test code was patched as well.


timj:
Adding a transform oracle for the joye_car headers found a real problem in the linux builds that do not use modern compiler settings and so were not as accurate. At some point maybe we force the builds to use `-march`.

PR #74, opened 2026-08-31
Prepare release notes for V9.4.0

timj:
v9.4.0 seems reasonable.

- [x] Blocked on #73 
- [x] Blocked on #37 (although I could be convinced to defer)
- [ ] Still need to convert the approved release notes to latex for the PDF docs #79 

## From Claude:

Summarize the changes made since V9.3.1 in ast.news and bump version.number to 9.4.0. The minor version is incremented because enforcing the documented Mapping immutability rule in every class can cause code that previously succeeded to report AST__IMMUT.

The new entries are grouped into behaviour changes, new features, bug fixes and testing/build changes. Related defect fixes are collected into a single entry per class, with nested lists for the FitsChan and KeyMap fixes, so that the volume of individual fixes in this release does not swamp the reader.

Generated with AI

Co-Authored-By: SLAC AI


embray:
I'm glad you are thinking about making a release because I had it on my todo list to start asking about the same.  libasdf-gwcs now depends on my fork of AST containing fixes that aren't upstream yet, and is statically linked in since a "normal" AST install as a shared library can't work yet anyways.

For my purposes I would really need

- #37 (as you already mentioned)
- A PR I haven't opened yet for adding support for another transform to YamlChan--I need to finish that.
- #66 
- #67 (ideally, though not strictly required)

I  could defer on:
- #51 though it would be good to have, and seems stalled.

That said, you were also talking on #37 about automated releases which is good, so maybe you are thinking of making a few more frequent? May fork/static lib approach is working fine for now so I don't feel the need to rush you on anything that isn't considered ready.

With AST now building as an independent library it will also be easier to package for conda and homebrew if you haven't already done that.  That would be really helpful for users, I think.

timj:
@embray thanks for the feedback. We already have conda-forge builds. Not homebrew. I am reminded that cmake install hasn't been tested at all but homebrew could use configure.

If I can solve the PDF problem (either by making it separately and uploading it or by writing a python prolat replacement) then I think we can pretty quickly turn around new releases as uploads to GitHub from within an action based on a tag. The ast.news stuff makes it a bit tricky to automatically add release notes. If we get in that position then I would have no objection to rapid releasing of features. I need to get the SEGV fix out for Rubin and I will try to use a more explicit semver model for version numbers. The one other wrinkle is that we have to hard code the new version into the source tree before applying the git tag but we can have a release check that checks that the version and the tag agree.

embray:
> The one other wrinkle is that we have to hard code the new version into the source tree before applying the git tag but we can have a release check that checks that the version and the tag agree

bumpver is pretty good for that.  I also put a tag/version string sanity check in my release workflow: https://github.com/asdf-format/libasdf/blob/56d24aa11b3013c362a485b25c2f51db35622d0e/.github/workflows/release.yml#L101

timj:
I think I'll try to make a quick release today without the YAMLchan fixes or #80 since we need that SEGV fix from #73

timj:
@embray I managed to make the release automatically by tagging. This is very good news. That means that I have no problem at all making new releases whenever you would like them.

@dsberry  I think it included a couple of old PRs in the release notes because those GitHub releases for 9.3.0 were marked pre-releases.

embray:
Too late for now (though could be added in retroactively later) but I was wondering if there shouldn't be some disclosure about AI tool use, as this was the first release to use it extensively.  This has been transparent in the commit and PR history.  I'm on the fence about whether such a disclosure is necessary, as even in the last few months it's rapidly become so much more commonplace that it seems almost redundant.  But I still do see projects that adopt LLM tools being charged, sometimes fairly, sometimes not, with becoming "slopified" or "vibe coded".  So I think it can still be useful to have a statement (albeit not overly defensive) outlining how LLMs were used, what benefit it added to the project (tons, I think), and what review process was followed to ensure that these were still high quality contributions.

Maybe also worth pointing out that it was used primarily for engineering tasks (build system, test coverage, bug fixes) and was not used to generate tons of novel scientific code that happened to pass generated tests (i.e. "slop").

timj:
@embray that's a good idea for the future. All the changes to AST itself done by LLMs have been tiny bug fixes.

I did experiment with using LLM for real algorithmic work in  #81 with the ChebyMap iterative inverse but I haven't got the faintest idea how to go about reviewing the code other than believing the LLM when it says that the iterative inverse does come up with the correct answer for a mapping that it knew the inverse of...

Issue #75, opened 2026-09-02
astLinearApprox can report a Mapping as linear when it departs from the returned fit by hundreds of times tol


timj:
This Claude-generated issue was created based on findings from #73 where it was discovered that `astLinearApprox` was reporting that a spline map was linear even though it had deviations that were larger than the requested tolerance.

## Summary

`astLinearApprox` (`src/mapping.c:6491`) decides linearity from a small, fixed set of sample positions expressed as constant fractions of the supplied box. Any structure in the Mapping that falls between those positions is invisible, however large it is and however well chosen the box is. Because the positions are fixed fractions rather than randomised, the failure is systematic and reproducible rather than a matter of luck: the same relative geometry defeats the test every time.

For a 2-input Mapping the function uses 17 positions in total:

- **4 fit positions**, at the centres of the box faces (`src/mapping.c:6690`). The returned fit is exact at these by construction.
- **13 test positions** (`npoint = 1 + 2*(ndim_in + 2^ndim_in)`, `src/mapping.c:6795` onward): one near the centre, one half way from the centre to each face centre, the four vertices, and one half way from the  centre to each vertex.

In normalised box coordinates those 17 positions are:
```
fit:   (0,0.5) (1,0.5) (0.5,0) (0.5,1)
test:  (0.51,0.51)
       (0.25,0.52) (0.76,0.52) (0.52,0.25) (0.52,0.76)
       (0,0) (1,0) (0,1) (1,1)
       (0.245,0.245) (0.765,0.245) (0.245,0.765) (0.765,0.765)
```
The gaps around `(0.35, 0.35)` and `(0.35, 0.65)` are roughly 0.15 of the box wide, which is comfortably large enough to hide a real feature.

The comment at `src/mapping.c:6845` says the offsets "introduce some small random offsets to break any regularity in the grid", but the offsets are the compile-time constants `0.48`/`0.52`/`0.49`/`0.51`. Nothing is randomised, so a caller cannot work around a miss by retrying.

## Reproducer

A bi-cubic `SplineMap` with 30 coefficients per axis — the resolution a real astrometric fit uses — set to the identity via the Greville abscissae, with a **single** coefficient moved. A single-coefficient bump has support `k/(n-k+1)` = `4/27` = 14.8% of the box, which fits in the gap.

```c
#include <stdio.h>
#include <math.h>
#include "ast.h"

#define KB 4
#define NB 30
#define TB (NB+KB)
#define L  1000.0

int main( void ) {
[...]

PR #78, opened 2026-09-08
Attempt to improve SIP linearity tests

timj:
Forked off #73.
Relevant issue: #75


timj:
For the record there was an unresolved comment on #73  that should probably be handled on this PR: https://github.com/Starlink/ast/pull/73/changes#r3960891749

GPT6 Astra says:

Currently, the code (src/fitschan.c:29067) measures the output displacement caused by a one-pixel step at the image centre. Jim’s concern is that a sharp local distortion could inflate that scale, making the subsequent linearity test too permissive. Review comment

I’d implement his “something else measured on large scales” suggestion using the existing 3×3 sample grid:

  1. Keep transforming those nine points through the lower mapping and polynomial to establish the upper mapping’s input bounds.
  2. Transform the opposite edge midpoints through the upper mapping too.
  3. Compute the output displacement across the full image width and height, divided by the corresponding pixel dimensions.
  4. Use those scales to calculate utol, then run the existing astLinearApprox test.

That removes the two extra one-pixel samples. We would need to preserve the original image dimensions before the bounding-box calculation overwrites ubnd, and reject invalid or zero scales.

Using the actual linear-fit coefficients is slightly more involved: astLinearApprox requires the tolerance before it returns a usable fit, and discards the coefficients if the test fails. It also fits only the upper mapping, whose inputs have already passed through the polynomial; its slopes alone are not necessarily output units per original pixel. The full-image measurement avoids both complications. AST’s own linear approximation already estimates slopes using opposite face centres, so this follows an established approach.

I’d also add a regression case in ast_tester/testfitschan.c:4757 where a large central derivative inflates the old tolerance enough to accept an unsuitable SIP approximation, and verify the revised calculation rejects it.

Overall: a few dozen lines in fitschan.c, a focused regression test, and no public API changes. This addresses the local-derivative concern; it still uses a sampled scale estimate rather than establishing a strict pixel-error bound everywhere.

PR #80, opened 2026-09-09
More claude bug reports and fixes

timj:


timj:
There are some additional bugs that need some thought (detail from Claude):


# Three defects where the fix is a design decision

While triaging the Rust porting team's `c-library-quirks.md` I confirmed a batch of real defects in `libast`. Most were mechanical and are already fixed on this branch:

| Commit | Fix |
|---|---|
| `6e75fb0b` | `astLoadXphMap` leaked the projection-type string it read |
| `5b4db859` | `Box::Simplify` overwrote the slice verdict instead of accumulating it |
| `f03bb2ca` | `Polygon::Simplify` tested only axis 1 of a transformed vertex for `AST__BAD` |
| `6c4e4052` | `PolyMap::Equal` never compared the inverse coefficients, and read out of bounds |
| `607c033f` | `PcdZoom` discarded an inverted neighbour ZoomMap's direction |
| `4c3c00fc` | `Ellipse::RegBaseMesh` measured the axis 2 bounds with axis 1's rule |

Three remaining ones are **not** mechanical: in each case the defect is clear but the correct behaviour after fixing it is a judgement call, and at least one of them changes results for existing callers. I would rather not pick unilaterally. Each section below gives the location, what goes wrong, a reproduction, and the specific question.

All line numbers and outputs are against this branch's HEAD (`433418ba`). Reproductions were run against a Debug build of this tree (`-DAST_ENABLE_SANITIZERS=ON`), **not** against a system or conda `libast`.

---

## 1. `PolyMap`'s linear `MapMerge` reduction changes a `ChebyMap`'s transformation

**Severity: this one returns a wrong answer from `astSimplify` today.** It is the only one of the three with a user-visible incorrect result rather than a missed or degraded one.

### Where

`chebymap.c` fills neither the `Simplify` nor the `MapMerge` vtab slot, and `polymap.c:2451` is the only place `MapMerge` is filled, so a ChebyMap's `MapMerge` is PolyMap's. Its replace-with-simpler half tests whether the forward transformation is linear and, if so, rebuilds it as a `MatrixMap` and a `ShiftMap` (`polymap.c:3759-3870`).

The test and the rebuild both read `pmap0->coeff_f` and `pmap0->power_f` directly. Nothing in that branch consults `scale_f` / `offset_f` — the per-input scale and offset a ChebyMap uses to map its input onto the interval the Chebyshev functions are defined over (`chebymap.c:2125-2126`):

```c
   new->scale_f[ i ]  = 2.0/( ubnd_f[ i ] - lbnd_f[ i ] );
   new->offset_f[ i ] = -( ubnd_f[ i ] + lbnd_f[ i ] )/( ubnd_f[ i ] - lbnd_f[ i ] );
```

For a PolyMap those coefficient arrays are the whole transformation, so the branch is correct. For a ChebyMap they are the coefficients of a Chebyshev series in the *normalised* input, and dropping the normalisation drops a scale and an offset. The two agree only when the bounds are exactly `[-1, 1]`, which makes the normalisation the identity.

### What it does

A one-input ChebyMap over bounds `[0, 10]` with the single term `2 * T_1`:

```
ChebyMap    f(0, 1, 10) = -2   -1.6   2
astSimplify              -> ZoomMap, Zoom = 2
simplified  f(0, 1, 10) =  0    2    20
```

`astSimplify` is documented to return a Mapping equivalent to the one supplied, so this is a defect rather than a convention. `scale_f = 0.2` and `offset_f = -1` for those bounds, so the map really computes `2*(0.2x - 1) = 0.4x - 2`, and the reduction keeps only the bare coefficient `2`.

### Why the fix is a decision

Two candidate fixes, and they are not equivalent.

**(a) Refuse the reduction for a ChebyMap.** Guard the branch with `!astIsAChebyMap( pmap0 )`. Minimal, obviously safe, and it is what the Rust port does (it has no `map_merge` for ChebyMap at all). The cost is that a linear ChebyMap is then never reduced, *including* over `[-1, 1]` where the existing reduction is already correct — so it is a small capability regression as well as a bug fix.

**(b) Fold the normalisation into the rebuild.** This is exact, and cheaper than it sounds. The existing linearity test (`polymap.c:3805-3830`) accepts a term only when every input power is 0, or exactly one input has power 1 and none exceeds 1. Since `T_0(z) = 1` and `T_1(z) = z` are the only Chebyshev polynomials of degree <= 1, a term the test accepts is *exactly* a term that is linear in the input as well as in the Chebyshev order. So with `z_j = scale_f[j]*x_j + offset_f[j]`:

```
   constant term (all powers 0):  offsets[iout] += c
   linear term (power 1 on j):    matrix[j + iout*nin] += c * scale_f[j]
                                  offsets[iout]        += c * offset_f[j]
```

For the example above that gives `matrix = 0.4`, `offset = -2`, i.e. `0.4x - 2` — the transformation the ChebyMap actually defines. One wrinkle: the current code *assigns* into `offsets[iout]` and `matrix[...]`, relying on earlier consolidation of equal-power terms to guarantee at most one of each. Under (b) a linear term also contributes to `offsets`, so those two writes have to accumulate.

I lean towards **(b)**: it fixes the bug without giving up the optimisation, and the arithmetic is short and provable. But it puts ChebyMap-specific knowledge into `polymap.c`, which is arguably the wrong place — the alternative shape is to give ChebyMap its own `MapMerge` that does the folding and delegates the rest.

### The missing piece: a ChebyMap has no iterative inverse

This is what makes the choice above matter more than it first appears.

A ChebyMap has three possible routes to an inverse, and only two of them exist:

1. **Explicit inverse coefficients** supplied to the constructor — works.
2. **`astPolyTran`**, which fits an inverse series by sampling — works; ChebyMap  overrides it (`chebymap.c:1193`, registered at `chebymap.c:1034`).
3. **The iterative inverse (`IterInverse`)** — **not implemented.**  `ChebyMap::GetIterInverse` (`chebymap.c:843`) returns 0 unconditionally, and  `polymap.c:6323-6327` documents that setting or clearing `IterInverse` on a ChebyMap has no  effect.

The reason is a storage gap, set out in the "To do" at the top of `chebymap.c:8-26`:

> Providing an iterative inverse requires the Jacobian to be defined.
>
>     for a PolyMap:   if y = C.x^n     then y' = n.C.x^n-1
>     for a ChebyMap:  if y = C.Tn(x)   then y' = n.C.Un-1(x)
>
> where Un-1(x) is the Chebyshev polynomial of the second kind [...] this would mean that
> ChebyMap needs to be able to represent Chebyshev polynomials of the second type. [...]
> Could use negative powers to indicate type 2, but PolyMap.StoreArrays objects to
> negative powers. [...] Simpler for the moment just to disable iterative inverses in
> ChebyMap.

The cost of that is already visible in this repository. `transform_oracle_overrides.txt` disables the round-trip check for `wcsconv/inputs/tnx-cheb.head` and `wcsconv/framesets/tnx-cheb-head.ast` under a comment marked "KNOWN ROUND-TRIP FAILURE -- INVESTIGATE": the TNX Chebyshev distortion builds a ChebyMap, so sky->pixel returns BAD for the whole fixture because no inverse exists. The answer to that INVESTIGATE note is this missing feature.

**Why this bears on (a) versus (b).** A forward-only ChebyMap has no inverse at all, and the linear reduction *creates* one — a MatrixMap and a ShiftMap are both analytically invertible. So the reduction is currently the only route by which a linear forward-only ChebyMap becomes invertible, and today that route hands back a *wrong* inverse:

```
forward-only ChebyMap, 2*T_1 over [0,10]:
  TranForward=1  TranInverse=0  IterInverse=0
  after astSet( "IterInverse=1" ) : IterInverse=0  TranInverse=0   (silently ignored)
  astSimplify -> ZoomMap : TranInverse=1           (an inverse appears)

  ChebyMap  fwd(0, 1, 10)     = -2   -1.6   2
  reduction inverse of those  = -1   -0.8   1      (should be 0, 1, 10)

same ChebyMap over [-1,1] : TranInverse=0 -> ZoomMap TranInverse=1
```

That reframes option (a). Refusing the reduction for a ChebyMap does not merely lose an optimisation — it removes the only inverse such a map can obtain without an explicit inverse or an `astPolyTran` call, including over `[-1, 1]` where the reduction is already correct. Option (b) fixes the value *and* keeps the gain: folding gives `0.4x - 2`, whose inverse recovers 0, 1 and 10 exactly.

On that evidence I think **(b) is clearly the right answer** and (a) should be rejected.

### A second decision in the same branch

Independently of the ChebyMap problem, this branch only examines the **forward** transformation and discards any explicit inverse. The source comment (`polymap.c:3759-3768`) is candid about it:

> We only check the original forward transformation since, if the simplificiation is
> possible, the inverse transformation will be defined automatically by the simpler
> Mappings (I suppose this could potentially throw away information if the forward and
> inverse transformations of the PolyMap are not a genuine inverse pair, but I can't
> immediately see why anyone would define an inverse transformation that was not a
> genuine inverse of the forward transformation).

This is reachable and observable. A PolyMap with forward `y = 2x` and inverse `x = 0.5y`, in series with the inverse-direction use of a PolyMap with forward `y = 2x` and inverse `x = 0.7y`, reduces to a `UnitMap` — the two are linearised, the explicit inverses are dropped, and the pair cancels. It was originally reported as a consequence of the `PolyMap::Equal` bug fixed in `6c4e4052`, but it is not: it still happens with `Equal` correct, because the linear route never reaches `Equal`.

### Questions

1. Does anyone disagree that (b) is the right fix, given that (a) would take away the only inverse a forward-only linear ChebyMap can get? If (b), should the folding live in  `polymap.c` or in a new `ChebyMap::MapMerge` that folds and delegates?
2. Is discarding an explicit, non-genuine inverse acceptable, as the comment assumes?  If not, this branch needs a `ncoeff_i == 0 || IterInverse` precondition, which would  narrow it noticeably. If yes, I will leave it and record it as intended. Note this is  less costly for a ChebyMap than it looks: the reduction replaces the explicit inverse  with an exact analytic one, so it is an upgrade unless the explicit inverse was not a  genuine inverse in the first place.
3. **Should the iterative inverse be implemented for ChebyMap?** It is a feature rather  than a defect, and the `chebymap.c` "To do" already sketches a design (negative powers  flagging second-kind polynomials, `StoreArrays` relaxed to store them and a virtual  function to verify them). It is the root cause of the two `tnx-cheb` round-trip  overrides. I have not scoped it; if it is wanted I would treat it as separate work  rather than folding it into this bug-fix branch.

---

## 2. `MergeInterval`'s PointList branch measures the uncertainty instead of the Interval

**Severity: this one also returns a Region that is not equivalent to its input** — see the transform comparison under "What it does". Two of the three issues in this document are therefore `astSimplify` correctness bugs, not cosmetic ones.

### Where

`interval.c:1456-1487` absorbs a neighbouring `PointList` into an `Interval`, producing a `PointList` in the combined base Frame. The Interval contributes one constant axis value per axis, which requires it to have zero width — the branch's own comment says so.

Two separate slips mean neither the test nor the values do that:

```c
         lbnd = astMalloc( sizeof( double )*(size_t) nax_this );
         ubnd = astMalloc( sizeof( double )*(size_t) nax_this );
         astRegBaseBox( this, lbnd, ubnd );                       /* the Interval    */

         lbnd_unc = astMalloc( sizeof( double )*(size_t) nax_this );
         ubnd_unc = astMalloc( sizeof( double )*(size_t) nax_this );
         bunc = astGetUncFrm( this, AST__BASE );
         astGetRegionBounds( bunc, lbnd, ubnd );                  /* <-- clobbers it */
         ...
               if( fabs( lbnd[ i ] - lbnd[ i ] ) >                /* <-- always 0.0  */
                   0.25*fabs( ubnd_unc[ i ] - lbnd_unc[ i ] ) ) { /* <-- never written */
```

1. `astGetRegionBounds` is handed `lbnd`/`ubnd` rather than the `lbnd_unc`/`ubnd_unc`  arrays declared for it at `interval.c:1467-1468`, so the Interval's own bounds — read  one statement earlier — are overwritten by the uncertainty Region's, and `lbnd_unc`/`ubnd_unc` are never written at all.
2. The width test subtracts `lbnd[i]` from itself, so its left side is always `0.0`, and its right side reads the two uninitialised arrays.

### What it does

The outcome is deterministic despite the uninitialised reads: `0.0 > 0.25*fabs(x)` is false for every `x`, `NaN` included, so `ok` stays 1 and the zero-width requirement never rejects anything. The constant axis values then become `0.5*( lbnd[i] + ubnd[i] )` over the *uncertainty's* bounds rather than the Interval's.

An Interval spanning `0..10` on axis 1 and unbounded above on axis 2 — nowhere near a point — merges anyway:

```
simplify( Prism( Interval[0..10, 0..open], PointList ) ) -> PointList
   extra axis values: 5, 0
```

Axis 1's `5` looks right only by coincidence: where no uncertainty was assigned, `astGetUncFrm` supplies the default one and `interval.c::GetDefUnc` centres that on the Interval's centre when every limit is finite. Axis 2 collapses to `0` because its limit is missing.

**The values are the lesser problem: the merged Region is not the Region it replaced.** A Region used as a Mapping maps a position to itself when the position is inside and to `AST__BAD` when it is outside, so the two can be compared directly. The Prism is `{ (x,y,z) : 0 <= x <= 10, y >= 0, z in {3,4} }`; the PointList it is replaced by holds just two points.

```
astEqual( prism, simplified ) = 0

  position          Prism      simplified
  (   2,    7, 3)   inside     outside      <-- inside the Prism, absent from the PointList
  ( 7.5,   12, 4)   inside     outside      <-- likewise
  (   0,    0, 3)   inside     outside      <-- likewise
  (  20,    1, 3)   outside    outside      agree
```

So this belongs in the same category as the ChebyMap defect above: `astSimplify` returns something that is not equivalent to its input. That is a stronger statement than "the constant axis values are wrong", and it raises the severity of this entry — I had it recorded as a values problem.

**One caveat, so the evidence is not overstated.** The merged PointList also reports its own two points, `(5,0,3)` and `(5,0,4)`, as outside itself. That is *not* caused by this branch. A hand-built 3-D PointList with the same two points, which never goes near `MergeInterval`, does the same:

```
3-D PointList, points differ on all three axes   own point -> inside
3-D PointList, constant on axes 1 and 2          own point -> OUTSIDE
3-D PointList, a single point                    own point -> OUTSIDE
2-D PointList, constant on axis 1                own point -> inside
```

It looks like a PointList's default uncertainty degenerates when the points have too little spread, and membership then fails everywhere. It interacts with this branch, because the merge always contributes a *constant* on each of the Interval's axes and so always produces points with no spread on those axes — but it is a separate question about PointList or about `GetDefUnc`, and I have not investigated it. The clean evidence above is the three positions that lie inside the Prism and nowhere near the merged points.

**Why nothing in the corpus catches this.** Worth knowing before deciding, because it affects what a fix has to come with:

- `simplify_<name>_astequal` compares the *simplified output* against the committed reference, not against the input, so for a positive fixture it checks reproducibility rather than equivalence. Only a *negative* fixture — whose reference is its own unsimplified `.map` — compares output against input, and there is no such fixture for this branch.
- The transform oracle's equivalence check (`compare_equiv` in `check_transform_oracle.c`) compares the live `.map` and `.simp` outputs but skips every point where the *recorded* `.map` and `.simp` disagreed at generation time, as being outside the shared valid domain. A fixture generated while this bug is present would therefore bake the divergence into the reference and exclude precisely the points that reveal it.

So the fixture has to be added **after** the fix, and as a negative fixture asserting the merge is refused. Adding one now would silently pin the wrong behaviour.

### Why the fix is a decision

The two slips are mechanical. What the corrected test should *be* is not:

- **The threshold contradicts its own comment.** The comment says "zero width means a width less than half the uncertainty on the axis"; the code uses `0.25`. Which is intended? (`0.25` of the full width is `0.5` of the half-width, so the two may have been meant to agree and the comment may be describing a half-width.)
- **Fixing it changes which pairs merge.** Any Interval/PointList pair that is not actually a point stops merging. The Rust port mirrors the current behaviour precisely because this determines corpus output, so this is a real behaviour change for downstream consumers, not just an internal tidy-up.
- **An axis with a missing limit has no defensible constant value.** Today it silently becomes `0`. I would expect the merge to be refused outright for such an axis, but that is an added restriction, not a restoration.

### Questions

1. Is the intended tolerance `0.25` of the full uncertainty width, or `0.5`?
2. Is a behaviour change to which Interval/PointList pairs merge acceptable in this
   release? If not, the alternative is to fix the uninitialised reads only — leaving the
   test vacuous — and to raise the semantics separately.
3. Should an unbounded axis refuse the merge rather than contributing `0`?

---

## 3. `CmpFrame::NormBox` can never widen a box, and errors when the components differ in size

### Where

`CmpFrame::NormBox` (`cmpframe.c:5830`) permutes the supplied bounds into internal order, splits them at the first component's axis count, and calls each component Frame's `astNormBox` with a probe Mapping that expands a component sub-position back into a full position before applying the caller's Mapping. Two things are wrong with the probes.

**The fill value defeats the test.** `pm2`, the probe for the first component (`cmpframe.c:5942-5946`), is built so its forward transformation copies the lower `naxes1` inputs to the same outputs and supplies `AST__BAD` for the others. Every caller of `astNormBox` — `Circle::RegBaseMesh` (`circle.c:1100`), `Ellipse::RegBaseMesh` (`ellipse.c:1189`), `Polygon::RegBaseBox` (`polygon.c:4317`) — passes a copy of itself with `Negated` cleared, and an un-negated Region maps a position carrying `AST__BAD` to `AST__BAD`. So the component's singularity test (`skyframe.c:7234`, `astTran2` on the two poles) always sees `AST__BAD` and always concludes the box holds neither pole. **A CmpFrame therefore never widens a box, for either component.**

**The second probe is malformed.** `pm3` (`cmpframe.c:5961-5962`) is

```c
      for( axis = 0; axis < naxes - naxes1; axis++ ) inperm[ axis ] = naxes1 + axis;
      pm3 = astPermMap( naxes1, inperm, naxes, NULL, NULL, "", status );
```

with `nin = naxes1` where the loop fills `naxes - naxes1` entries, and with a null `outperm`, whose forward transformation copies input *i* to output *i* while *i < nin* and supplies `AST__BAD` after that. The second component's sub-position therefore arrives on the *first* component's internal axes. The inverse, built from `inperm`, extracts the correct axes, so only the forward direction — the one the probe uses — is affected.

When the second component's axis count differs from the first's, the arity check in `astTran2` rejects the probe outright with `AST__NCPIN`. `SkyFrame::NormBox` does not check status after that call and reads the uninitialised `xo` / `yo` arrays it passed (`skyframe.c:7237`), widening the box from whatever they hold.

### What it does

A Circle 0.3 rad from the north pole with radius 0.5, so the pole is comfortably inside:

```
plain SkyFrame          : lon 0..6.28318530717959     lat ..1.5707963267949   (= pi/2)
CmpFrame(SkyFrame, 1-D) : lon 0..6.14787543600455     lat ..1.53631396098367
CmpFrame(1-D, SkyFrame) : status = AST__NCPIN, "Bad number of input coordinate values"
CmpFrame(SkyFrame, SkyFrame) 2+2:
                          comp1 lat ..1.570224009082  comp2 lat ..1.570224009082
```

The plain SkyFrame widens correctly. The CmpFrame returns the raw mesh bounds. The 2+2 case shows that making the component sizes equal only gets past the arity check — the box is still not widened, because `pm2`'s `AST__BAD` fill defeats the test regardless.

### Why the fix is a decision

The `pm3` arity and `outperm` are mechanical, and the missing `astOK` check in `skyframe.c` should go in unconditionally — reading uninitialised memory after an error is indefensible however the rest is resolved.

The fill value is the hard part. The probe has to supply *something* on the other component's axes, and per `astNormBox`'s own contract the Mapping must "leave an input position unchanged if the point is inside the box, and set all bad if the point is outside" — so the value has to be **inside the Region** or the test answers the wrong question.

- For an **axis-aligned box** the natural choice is the centre of the supplied bounds on the other component's axes, and it is provably right: the box is a product of per-axis ranges, so if the pole lies inside the sky sub-box then the full position (pole, other-centre) lies inside the full box.
- But the caller passes a **Region**, not a box — a Circle, Ellipse or Polygon on the CmpFrame. A Region on a CmpFrame is not a product of its components, so "does this Region contain the pole?" genuinely cannot be decomposed per component. Any single fill  value can give a false negative (the pole is in the Region at some other-axis value, just not the one we picked).

So a correct-in-general fix may not exist in this shape, and the realistic options are:

1. **Fill with the box centre on the other axes.** Cheap, right for a box-like Region, and strictly better than today (which is always a false negative). Can still miss.
2. **Fill and retry over a small set of positions** (centre plus the bounds' corners on the other axes), widening if any of them reports the singularity inside. More  expensive, fewer false negatives, still not exact.
3. **Test the singularity against the full Region directly**, without decomposing — i.e. have `CmpFrame::NormBox` ask each component *which* singular positions exist and then test the assembled full positions itself, rather than handing each component a probe. Closest to correct, largest change.
4. **Make the current behaviour explicit**: document that a CmpFrame does not widen for singularities, fix only `pm3` and the uninitialised read. Honest and safe, fixes no geometry.

Note that whichever is chosen, options 1-3 **change the bounds reported for every CmpFrame containing a SkyFrame** whose box holds a pole, which affects `astGetRegionBounds` and everything downstream of the mesh (`astRegPins`, `astMask`, Region simplification).

### Questions

1. Which of 1-4? I lean towards 1 as an incremental improvement with 4's documentation alongside it, but 3 is the only one that is actually right, and it may be worth doing properly if anyone relies on CmpFrame region bounds.
2. Is a change to CmpFrame region bounds acceptable now, or should it wait for a release where downstream consumers expect geometry changes?

---

## What I will do without guidance

If there is no strong opinion I will take the conservative reading throughout: (1b) for ChebyMap, since it is a plain bug with a provable fix and the alternative would remove an inverse; the ChebyMap iterative inverse left alone as separate work; for (2) the uninitialised reads plus `0.25` retained as written, with the vacuous test left vacuous and flagged; and for (3) only `pm3`, the `astOK` check, and documentation. That fixes the wrong answer, removes the undefined behaviour, and leaves every semantic change for a deliberate decision.

Each fix will get both a generated fixture and an explicit C test, as the rest of this branch does — except where one is impossible, which I will say so in the commit message.

timj:
@dsberry do you have any thoughts on the

> The comment says "zero width means a width less than half the uncertainty on the axis"; the code uses 0.25. Which is intended?

question where the comment says 0.5 and the code says 0.25?

I'm trying to deal with the ChebyMap issue in #81

timj:
@dsberry as requested I have reverted the IsSimp change so that the test simplification for the clone does now affect the original.

PR #81, opened 2026-09-09
Add ChebyMap iterative inverse

timj:


timj:
An analysis of the TPN ChebyMap issue can be found in #84 and makes the argument that it's a seed issue and also that the test header has an inverse with some bad properties. No further analysis is needed on this ticket.

PR #89, opened 2026-09-16
DM-56110: Fixes found from astshim unit tests

timj:
* Fix astNormUnit so that it returns the unit unchanged if it can't parse it (which is the documented behavior that never worked and only became obvious now that  15182b5a fixed a bug (but had no test).
* Fixed `astEQUALS` so that we combine numbers in the right order.

Also make `cmake` install match what autotools does.


src/frame.c
```diff
@@ -13038,9 +13049,104 @@ MAKE_TEST(Unit)
...
-/* This simply provides an interface to the Axis methods for accessing
-   the NormUnit string. */
-MAKE_GET(NormUnit,const char *,NULL,0,0)
+static const char *NormUnitString( AstFrame *this, int axis, int *status ) {
+/*
+*  Name:
+*     NormUnitString
+
+*  Purpose:
+*     Return the normalised form of the Unit value a Frame reports for an
```
timj, line 13058: @dsberry  I'm basing this change on the wording for `NormUnit`:

>  If no simplification can be performed, the value of the NormUnit attribute will equal that of the Unit attribute.

and so if the unit is `hh:mm:ss.sss` you get that back instead of an error about parsing.

